Issue #6, opened 2026-06-11
Helm chart: sqlite.persistence.existingClaim to use a pre-created PVC instead of volumeClaimTemplates


joryirving:
## Context

Same deployment as #5: memini 0.0.4 on Kubernetes via the bundled chart, sqlite backend. My PVCs are managed externally by VolSync — the claim is pre-created with a `dataSourceRef` pointing at a `ReplicationDestination`, so the volume is restored from backup before the workload mounts it.

## Problem

The sqlite StatefulSet hard-codes a `volumeClaimTemplates` entry named `data`; there's no way to point the chart at an existing claim. It can be made to work today by relying on StatefulSet PVC adoption — pre-create the claim as `data-<fullname>-0` (pinned via `fullnameOverride`) and the controller adopts it instead of provisioning. But that's implicit and has a sharp edge: if the PVC is ever deleted while the StatefulSet exists, the controller instantly recreates a blank one without the `dataSourceRef`, silently skipping restore. Anyone doing restore-from-backup has to remember to delete the StatefulSet first.

## Proposal

An `existingClaim` option under `sqlite.persistence`, mirroring what most charts (and e.g. bjw-s app-template) do:

```yaml
sqlite:
  persistence:
    enabled: true
    existingClaim: ""   # when set: mount this PVC instead of rendering volumeClaimTemplates
    size: 5Gi
    storageClass: ""
    accessModes: [ReadWriteOnce]
```

When `existingClaim` is set, the StatefulSet renders a plain pod `volumes:` entry:

```yaml
volumes:
  - name: data
    persistentVolumeClaim:
      claimName: {{ .Values.sqlite.persistence.existingClaim }}
```

and skips `volumeClaimTemplates` entirely (a StatefulSet with no claim templates is fine, and the single-replica sqlite mode doesn't need per-replica claims anyway). `enabled: false` → emptyDir stays as-is.

Happy to send a PR — it's a small change to `statefulset.yaml.tpl` + values/schema.

PR #8, opened 2026-06-11
helm chart: sqlite.persistence.existingClaim for pre-created PVCs

joryirving:
Implements #6.

When `sqlite.persistence.existingClaim` is set, the StatefulSet mounts the named PVC as a plain pod volume and skips `volumeClaimTemplates` entirely (single-replica sqlite doesn't need per-replica claims). `enabled: false` → emptyDir is unchanged, and the default path still renders the volumeClaimTemplate exactly as before.

Motivation: externally managed claims — in my case a VolSync-managed PVC created with a `dataSourceRef` pointing at a `ReplicationDestination` so the volume is restored from backup before the workload mounts it. That works today only via implicit StatefulSet PVC adoption (pre-creating `data-<fullname>-0`), which has a sharp edge: delete the PVC while the StatefulSet exists and the controller recreates a blank claim without the dataSourceRef, silently skipping restore.

README/values.schema.json regenerated with the CI-pinned helm-docs 1.14.2 / helm-schema 0.23.4 (also picks up the two missing `ui` README rows). Verified with `helm lint` and `helm template` for all three modes: default (volumeClaimTemplates byte-identical to before), `existingClaim=my-claim` (plain volume, no claim templates), `enabled=false` (emptyDir).

Note: this and the rerank PR both regenerate README/schema, so whichever merges second needs a trivial regen rebase.

### DIFF
--- a/charts/memini/README.md
+++ b/charts/memini/README.md
@@ -65,8 +65,11 @@ helm install memini ./charts/memini \
 | service.type | string | `"ClusterIP"` |  |
 | serviceAccount.create | bool | `true` | Create a ServiceAccount |
 | serviceAccount.name | string | `""` | ServiceAccount name; generated when empty |
-| sqlite | object | `{"persistence":{"accessModes":["ReadWriteOnce"],"enabled":true,"size":"5Gi","storageClass":""}}` | sqlite backend persistence (StatefulSet volumeClaimTemplate) |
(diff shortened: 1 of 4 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-06-11 16:08 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/13#issuecomment-7763)_

/sync

eleboucher:
> _Originally by **bot-owl** on 2026-06-11 16:38 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/13#issuecomment-7774)_

Promoted to #16. Future updates will sync to that PR.

Issue #12, opened 2026-06-11
OpenClaw extension: option to skip sessions with no agent identity (cron/heartbeat)


joryirving:
## Context

Following the per-agent namespace work in #4. Running the OpenClaw extension with `namespace_per_agent: true` across several agents (alice/bob/carol), I noticed a stray `openclaw` namespace filling with episodic memories.

## Cause

The gateway fires `agent_end` (and `before_agent_start`) for **cron and heartbeat sessions** that carry no agent identity — none of the `agentIdentity()` candidate fields are populated. `effectiveNamespace()` then falls back to the static base namespace (default `openclaw`), so automation runs accumulate episodic memory in a shared bucket, polluting recall.

These sessions can't be redirected to a per-agent namespace (the event has no id to key on), so the useful behavior is to **not write them at all**.

## Proposal

An opt-in `skip_without_agent` boolean. In `namespace_per_agent` mode, when no agent id resolves and the flag is set, `effectiveNamespace()` returns `null`, and both hooks skip the operation (no recall, no write, no fallback namespace). Default behavior (fall back to base namespace) is unchanged.

PR attached — small change to `plugin.mjs` (config + `effectiveNamespace` + two hook guards) plus two test cases; existing tests still pass (8/8).

Issue #4, opened 2026-06-11
OpenClaw extension: per-agent namespace support


joryirving:
## Context

I'm migrating a multi-agent setup from agentmemory to memini (great project, thanks — the agentmemory importer and the OpenClaw integration made this an easy sell). My OpenClaw gateway runs several agents (e.g. `alice`, `bob`, `carol`), and I'd like each agent to have its own memini namespace so their memories don't cross-pollinate.

## Problem

The OpenClaw extension (`integrations/openclaw/plugin/plugin.mjs`) resolves a single static `namespace` from plugin config at `register()` time:

```js
namespace: api.pluginConfig?.namespace || DEFAULT_NAMESPACE,
```

The `before_agent_start` / `agent_end` hooks then send that one `X-Memini-Namespace` header for every agent in the gateway, so all agents share one memory namespace. The README documents this as intentional for the shared case, but there's currently no way to opt into per-agent isolation.

## Proposal

Derive the namespace from the agent identity when configured to do so, e.g.:

```jsonc
"memini": {
  "enabled": true,
  "config": {
    "base_url": "http://memini:8080",
    "namespace": "openclaw",          // current behavior, default
    "namespace_per_agent": true,       // new: use agent id as namespace
    "namespace_template": "{agent}"  // or a template, e.g. "openclaw-{agent}"
  }
}
```

If the `before_agent_start` / `agent_end` events expose an agent id/name, the client could resolve the header per call instead of once at register time. Falling back to the static `namespace` when the agent id is unavailable would keep current behavior intact.

Since memini already multi-tenants cleanly via `X-Memini-Namespace`, this is purely an extension-side change — happy to test a branch against a live multi-agent gateway, or send a PR if you'd prefer and can confirm the agent id is available on those events.

PR #15, opened 2026-06-12
openclaw plugin: resolve agent identity from hook ctx and session keys

joryirving:
Fixes per-agent namespacing (#4) actually working against a real OpenClaw gateway.

**Problem:** OpenClaw's `agent_end` event payload is `{type, messages}` — none of the identity fields `agentIdentity()` checks exist on it (verified against the gateway dist: `emit({ type: "agent_end", messages: event.messages })`). So with `namespace_per_agent: true`, identity never resolves and **every** write lands in the static fallback namespace. Multi-agent gateways get one co-mingled namespace — the thing #4 was meant to prevent.

**Fix:** the gateway invokes extension handlers as `handler(event, ctx)`, and ctx carries agent-scoped session keys (`agent:<id>:...`, e.g. `agent:carol:cron:daily`). Accept `ctx` in both hooks; resolve identity from `ctx.agentId`/`agent_id` first, then by parsing the `agent:<id>` segment from ctx/event sessionKey/sessionId/runId. (Same mechanism the agentmemory OpenClaw plugin used, which demonstrably produced per-agent attribution.)

**Behavior change:** raw session ids are no longer treated as identities — a bare session UUID isn't an agent, and using it fragments memory into per-session namespaces. Session ids now contribute only via the `agent:<id>` parse; one existing test updated accordingly.

Composes with `skip_without_agent` (#12/#13) to give exact semantics: agent sessions — including their crons/heartbeats — route to the agent's namespace; gateway-level unattributable sessions are skipped.

`node --test`: 10/10 (3 new cases incl. ctx resolution and the cron/heartbeat split).

### DIFF
--- a/integrations/openclaw/plugin/openclaw.plugin.json
+++ b/integrations/openclaw/plugin/openclaw.plugin.json
@@ -13,6 +13,7 @@
       "namespace": { "type": "string" },
       "namespace_per_agent": { "type": "boolean" },
       "namespace_template": { "type": "string" },
+      "skip_without_agent": { "type": "boolean" },
       "fallback_on_error": { "type": "boolean" },
       "timeout_ms": { "type": "number" }
     }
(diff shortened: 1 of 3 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-06-12 14:22 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/23#issuecomment-7990)_

/sync

eleboucher:
> _Originally by **bot-owl** on 2026-06-12 14:23 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/23#issuecomment-7993)_

Promoted to #25. Future updates will sync to that PR.

eleboucher:
Promoted to the canonical Forgejo repository (PR #25), which is the source of truth. Further review and merging happen there.

PR #23, opened 2026-06-17
fix(openclaw): strip OpenClaw runtime preambles from captured turns

### LINKED ISSUE #22: OpenClaw plugin: captures OpenClaw runtime preambles (Conversation info / Subagent Context) as memory → recall pollution
## Summary

The OpenClaw plugin's `agent_end` capture stores the raw last user/assistant text, which includes OpenClaw's **runtime preambles** that aren't memory content. The dominant one is the `Conversation info (untrusted metadata)` block OpenClaw prepends to user turns; `[Subagent Context] …` framing is another. These accumulate as episodic memories and — because they're templated/repetitive — recall at **high** relevance scores, crowding out real memories during `before_prompt_build` injection.

## What gets captured (verbatim)

```
User: Conversation info (untrusted metadata):
```json
{ "chat_id": "...", "message_id": "...", "sender_id": "...", "sender": "..." }
```
<the actual user message follows here>
```

```
User: [Subagent Context] You are running as a subagent (depth 1/1). Results auto-announce to ...
```

In one busy agent namespace, **~50% of recently-captured episodic memories were these preambles** (the single largest class being the `Conversation info` block), recalling at **0.80–0.94** similarity — so a `min_score` floor doesn't filter them.

## Why `skip_system_turns` doesn't cover it

`skip_system_turns` (great for cron/heartbeat/scheduled) keys off system "kinds" — session-key segments and `[cron:]`/`[OpenClaw heartbeat poll]` prefixes. But the `Conversation info` block rides on a **genuine user turn** (OpenClaw just prepends untrusted chat metadata), so it isn't a system kind, and skipping the whole turn would drop the real message.

## Suggested fix

In the capture path, before storing, **strip OpenClaw's known runtime preambles** from the user text rather than skipping the turn:
- Drop a leading `Conversation info (untrusted metadata): { … }` block (it's explicitly *untrusted metadata*, not memory) and keep the actual message that follows.
- Strip or skip `[Subagent Context] …` framing.
- Optionally make the set of strippable preambles configurable.

This preserves the real conversation while dropping the plumbing. Without it, the plugin re-pollutes recall with metadata blocks even with `skip_system_turns` enabled, and operators have to periodically purge by hand.

Distinct from #19 (cron/heartbeat *kinds*) and #21 (injection budgeting) — this is about stripping non-memory preambles from otherwise-legitimate captured turns.

### PR DESCRIPTION
**Pull request by joryirving**
Fixes #22.

## Problem

OpenClaw injects runtime plumbing into the user turn the model (and `agent_end`) sees:

- one or more `<Label> (untrusted metadata):` blocks (chat info, sender info), each followed by a fenced JSON object, and
- a `[Subagent Context] You are running as a subagent …` preamble on subagent task delegations.

`agent_end` captured these verbatim as episodic memories. Because they're templated and repeat across turns, they recall at **high** similarity (0.8–0.94 in practice) and crowd out real memories during `before_prompt_build` injection. A `min_score` floor can't filter them (the scores are high), and `skip_system_turns` doesn't catch them — the metadata block rides on a *genuine* user turn, so it isn't a cron/heartbeat kind.

What got stored, e.g.:

```
User: Conversation info (untrusted metadata):
```json
{ "chat_id": "...", "sender_id": "...", "sender": "..." }
```

Sender (untrusted metadata):
```json
{ "name": "..." }
```

<the actual user message>
```

## Change

In the `agent_end` capture path, after the existing system-turn check (so cron/heartbeat detection still runs on the raw text):

- `stripRuntimePreambles()` removes leading `… (untrusted metadata):` + fenced-JSON blocks (handles stacked blocks), keeping the real message.
- Skip capture when the turn is a pure `[Subagent Context]` delegation (framing, not conversation), or when stripping leaves nothing.

No config/schema change — the metadata blocks are unambiguously OpenClaw plumbing, so stripping is unconditional.

## Tests

Added 4 unit tests for `stripRuntimePreambles` (single block, stacked blocks, plain message untouched, metadata-only → empty). Full suite: **45 passing** (`node --test`, deps installed).

### DIFF
--- a/integrations/openclaw/plugin/plugin.mjs
+++ b/integrations/openclaw/plugin/plugin.mjs
@@ -262,6 +262,24 @@ function lastTextByRole(messages, role) {
   return "";
 }
 
+// OpenClaw prepends runtime plumbing to the user turn that the model sees:
+// one or more "<Label> (untrusted metadata):" blocks (chat/sender info), each
+// followed by a fenced JSON object. That metadata is explicitly untrusted and
+// is not memory — captured verbatim it dominates a namespace and recalls at
+// high similarity (it's templated), crowding out real memories. Strip the
+// leading metadata blocks and keep the actual message that follows.
+const UNTRUSTED_METADATA_BLOCK =
+  /^\s*[^\n]*\(untrusted metadata\):\s*```(?:json)?\s*[\s\S]*?```\s*/;
+
+export function stripRuntimePreambles(text) {
+  if (typeof text !== "string") return text;
+  let out = text;
+  while (UNTRUSTED_METADATA_BLOCK.test(out)) {
+    out = out.replace(UNTRUSTED_METADATA_BLOCK, "");
+  }
+  return out.trim();
+}
+
 // DEFAULT_RECALL_LABELS mirrors the opencode plugin's labels toggle. Empty by
 // default — when labels are off, formatResults renders the legacy
 // "(tier) text" line so existing callers see identical output.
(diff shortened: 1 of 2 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-06-17 16:04 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/43#issuecomment-8738)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #44), which is the source of truth. Further review and merging happen there.

PR #27, opened 2026-06-18
fix(openclaw): register hooks via api.on so a throwing registerHook can't abort register() (#26)

### LINKED ISSUE #26: openclaw plugin 0.4.8: register() throws 'hook registration missing name' → memini never loads
## Summary

The rewritten (TypeScript) OpenClaw plugin in **0.4.8** installs cleanly (the `dist/` packaging fix from 0.4.8 works — `.clawver` updates, `dist/index.js` is present), but **fails `register()` at gateway startup**, so memini never loads:

```
[plugins] loading memini from .../extensions/memini/dist/index.js
[plugins] memini failed during register from .../dist/index.js: Error: hook registration missing name
[plugins] 1 plugin(s) failed to initialize (register: memini)
[gateway] plugin failed during register: Error: hook registration missing name (plugin=memini)
```

Result: the gateway starts but with memini absent from the loaded-plugin set — no recall, no capture.

## Likely cause

`register()` registers hooks via both APIs (src/index.ts):

```ts
api.registerHook("before_prompt_build", recallHandler);
api.on?.("before_prompt_build", recallHandler);
// ...
api.registerHook("agent_end", captureHandler);
api.on?.("agent_end", captureHandler);
```

Production OpenClaw's `api.registerHook` appears to reject the `(name, handler)` positional form with "hook registration missing name" (it likely wants a `{ name, handler }` object, or a handler carrying an explicit name). The throw aborts `register()` before the `api.on?.` fallback can take effect. The previous `plugin.mjs` (0.4.6) used `api.on("before_prompt_build", handler)` and loads fine, so the `registerHook` call is the regression.

## Suggested fix

Match production OpenClaw's `registerHook` signature (object form / explicit name), or guard `registerHook` so a signature mismatch doesn't abort registration (fall through to `api.on`). The hook bodies themselves (incl. the runtime-preamble strip) are correct — only the registration wiring is broken.

## Impact

0.4.7 (no `dist/`) and 0.4.8 (this) are both unusable for OpenClaw; the last loadable version is 0.4.6 (pre-rewrite `plugin.mjs`), which lacks the strip fix and the `recall_limit`/`recall_min_score` config.

joryirving:
## Problem

Fixes #26. The TypeScript OpenClaw plugin (0.4.7/0.4.8) installs cleanly but **fails `register()` at gateway startup**, so memini never loads — no recall, no capture:

```
[plugins] loading memini from .../extensions/memini/dist/index.js
[plugins] memini failed during register: Error: hook registration missing name
[plugins] 1 plugin(s) failed to initialize (register: memini)
```

`register()` calls `api.registerHook("before_prompt_build", recallHandler)` (positional). Production OpenClaw rejects that form with `hook registration missing name`, and the throw aborts `register()` **before** the `api.on?.(...)` fallback on the next line can run — so neither hook is wired and the plugin is dropped. The previous `plugin.mjs` (0.4.6) registered via `api.on(...)` and loads fine, so `api.on` is present in production; only the `registerHook` call is the regression.

## Fix

Register through a small `addHook` helper that:
- prefers `api.on` (the surface that actually loads in production), and
- falls back to `registerHook`'s **object** form (`{ name, handler }`) only when `api.on` is unavailable, wrapped so a signature mismatch can never abort `register()`.

The hook bodies (incl. the runtime-preamble strip and recall gating) are unchanged — only the registration wiring.

## Test

Adds a regression test to `test/regression.test.mjs` with a mock whose `registerHook` **throws** `hook registration missing name` (replicating production) while `on` captures. It asserts `register()` doesn't throw and both hooks are wired. The test **fails against the current code** and passes with this change:

```
# old registration:  ✖ register survives a throwing api.registerHook ...  (1 fail)
# this PR:            ✔ ... 46 pass / 0 fail
```

`npm run build`, `npm run typecheck`, and `npm test` all pass.

### DIFF
--- a/integrations/openclaw/plugin/src/index.ts
+++ b/integrations/openclaw/plugin/src/index.ts
@@ -623,10 +623,25 @@ const plugin: {
       if (!block) return;
       return { prependContext: `Relevant long-term memory from memini:\n${block}` };
     };
-    api.registerHook("before_prompt_build", recallHandler);
-    // api.on is the legacy hook surface; some test/runtime mocks still use it.
-    // Production OpenClaw only has api.registerHook.
-    api.on?.("before_prompt_build", recallHandler);
+    // Register on whichever hook surface this OpenClaw build exposes. api.on is
+    // the surface the prior plugin.mjs used and is present in current
+    // production; a positional api.registerHook(name, handler) is rejected with
+    // "hook registration missing name" and would abort register(), dropping the
+    // memory hooks entirely (eleboucher/memini#26). Prefer api.on; fall back to
+    // registerHook's object form only when api.on is unavailable, and never let
+    // it throw.
+    const addHook = (name: string, handler: any) => {
+      if (typeof api.on === "function") {
+        api.on(name, handler);
+        return;
+      }
+      try {
+        api.registerHook?.({ name, handler });
+      } catch {
+        /* unknown registerHook signature; nothing else to try */
+      }
+    };
+    addHook("before_prompt_build", recallHandler);
 
     const captureHandler = async (event: any, ctx: any) => {
       if (!cfg.enabled || !Array.isArray(event.messages)) return;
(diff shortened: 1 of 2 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-06-18 04:39 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/49#issuecomment-8826)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #50), which is the source of truth. Further review and merging happen there.

Issue #21, opened 2026-06-17
make automatic memory injection budgeted, curated, and user-controllable


joryirving:
### Proposal: make automatic memory injection budgeted, curated, and user-controllable

I’m trialing Memini with OpenClaw using per-agent namespaces. The backend design feels strong: tiered memory, hybrid retrieval, reranking, deduplication, MCP tools, and session digests are all the right primitives.

The pain point is the automatic injection layer. In practice, the plugin can inject too many memories into a session or before tool use, especially for smaller/fast models. The result is that memory becomes noisy context rather than helpful continuity.

I’d like to propose treating “automatic injection policy” as a first-class feature, separate from storage and recall.

#### Desired behavior

Memini should behave less like “retrieve N related memories” and more like a small top-of-mind briefing:

* Store lots of memories.
* Auto-inject only a tiny curated subset.
* Keep the rest searchable through MCP tools.
* Label injected memories with trust/freshness/tier.
* Make injection budgets configurable per hook and integration.

#### Suggested config

```yaml
memory_injection:
  session_start:
    enabled: true
    max_total_items: 4
    max_tokens: 500
    sections:
      pinned: 1
      semantic: 2
      procedural: 1
      recent: 0

  pre_tool_use:
    enabled: true
    max_total_items: 1
    max_tokens: 180
    min_score: 0.65
    tools: ["Read", "Edit", "Write"]

  formatting:
    include_tier: true
    include_confidence: true
    include_age: true
    include_reason: true
```

#### Suggested product concept: top-of-mind memories

Instead of auto-injecting raw search results, Memini could maintain a small curated “top-of-mind” set per namespace. These would be the durable preferences, conventions, active goals, and confirmed high-value facts that are safe to inject by default.

Everything else should remain available through `memory_recall`, but not automatically inserted into every session.

#### Why this matters

For agent systems, especially smaller or faster models, injected memory has a lot of authority. Too much memory causes confusion, stale assumptions, and cross-task pollution. The ideal behavior is not “more memory,” but “smaller, more trustworthy memory by default.”

PR #28, opened 2026-06-18
fix(openclaw): don't re-inject the same memories on every step of a turn (#21)

joryirving:
## Problem

Re #21 (closed). `recall_limit`/`recall_min_score` gave a single global cap + optional floor, but the noisiest part of automatic injection is still there: **the same memories get re-injected on every step of a turn.**

`before_prompt_build` fires on every step (each tool call), and an unchanged user query recalls the same top memories each time — so the `Relevant long-term memory from memini:` block is re-prepended on every step. A multi-tool turn ends up with the same memories stacked repeatedly, drowning the prompt — especially painful for smaller/fast models. This is the "spray" half of #21's complaint, independent of the per-hook budget design.

## Fix

Track, per session, which memory ids have already been surfaced, and drop repeats on later steps:

- First injection of a turn is **unchanged** (backward-compatible — same memories, same formatting).
- Later steps only inject memories not already shown this session; genuinely new matches still surface.
- The session→ids map is bounded (200 sessions, oldest evicted) so a long-lived gateway can't leak.

No new config and no default changes — purely removes redundant re-injection. Complements `recall_min_score` (quality floor) and `recall_limit` (per-call cap) rather than replacing them.

## Test

Adds a regression test: two `before_prompt_build` calls in the same session with the same query → the second injects nothing (already shown); a different session still sees them. It **fails without the dedup** and passes with it. `npm run build`, `npm run typecheck`, `npm test` (47 + 25) all green.

## Note / possible follow-up

This doesn't implement #21's full per-hook budget (`session_start` vs `pre_tool_use`, token budgets, tier sections) — just the highest-impact, lowest-risk slice. Happy to take it further if you'd like the budgeted-injection layer too.

### DIFF
--- a/integrations/openclaw/plugin/src/index.ts
+++ b/integrations/openclaw/plugin/src/index.ts
@@ -604,6 +604,28 @@ const plugin: {
       });
     }
 
+    // OpenClaw fires before_prompt_build on every step of a turn (each tool
+    // call), and an unchanged query recalls the same top memories — so the same
+    // "long-term memory" block is re-injected on every step, drowning the prompt
+    // (eleboucher/memini#21). Track what each session has already been shown and
+    // drop repeats on later steps; genuinely new matches still surface. Bounded
+    // so the map can't grow without limit across long-lived gateways.
+    const injectedBySession = new Map<string, Set<string>>();
+    const MAX_TRACKED_SESSIONS = 200;
+    const rememberInjected = (session: string, ids: string[]) => {
+      let seen = injectedBySession.get(session);
+      if (!seen) {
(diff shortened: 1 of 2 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-06-18 21:02 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/51#issuecomment-8965)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #52), which is the source of truth. Further review and merging happen there.

PR #29, opened 2026-06-21
feat(openclaw): reserve recall slots for consolidated semantic/procedural memory (#21)

joryirving:
## Problem

Re #21. Recall pulls the top `recall_limit` memories by relevance across **all tiers**. But episodic per-turn captures outnumber the LLM-distilled tiers by ~20:1 in practice (one real namespace: **299 episodic / 16 semantic / 6 procedural**), so the top matches are almost always conversational chatter ("how far did the cron get", "keep going") and the consolidated facts/rules — exactly the durable knowledge worth surfacing — rarely make the cut. `recall_limit` and `recall_max_tokens` bound *volume* but can't bias *what kind* of memory wins a slot.

## Fix

Add `recall_semantic_reserve` (default `0` = unchanged). When `> 0`, recall over-fetches a candidate pool and guarantees up to N of the `recall_limit` slots go to **durable** memories (semantic + procedural), filling the remaining slots by relevance. Output preserves relevance order.

- **Back-compat:** `0` reserves nothing — byte-identical behavior to today.
- **No server change:** results already carry `memory.tier`; composition is client-side. (`/v1/search` has no tier filter, so the pool is over-fetched then composed down.)
- Complements the existing knobs: `recall_limit` (count) + `recall_max_tokens` (token budget) + `recall_semantic_reserve` (tier floor) — the per-tier "sections" idea from #21, scoped to one small, optional knob rather than a full injection-policy DSL.

`reserveDurableTiers()` is exported and unit-tested directly.

## Tests

- `resolveConfig` default (`0`) + honored value.
- `reserveDurableTiers`: reserve 0 = plain top-N; reserve 1/2 promote semantic/procedural over episodic while keeping relevance order; reserve capped at limit; small-pool passthrough.

`npm run build`, `npm run typecheck`, `npm test` (49 + 27) all green.

## Note

This is a deliberately small slice of #21 — a tier floor, not the full per-hook budget/sections config. Happy to extend toward the larger design if you'd prefer it server-side (e.g. a tier weight in fusion, which would also let the reserve work without over-fetching).

### DIFF
--- a/integrations/openclaw/plugin/openclaw.plugin.json
+++ b/integrations/openclaw/plugin/openclaw.plugin.json
@@ -28,7 +28,8 @@
       "timeout_ms": { "type": "number" },
       "expose_tools": { "type": "boolean" },
       "recall_limit": { "type": "number" },
-      "recall_max_tokens": { "type": "number" }
+      "recall_max_tokens": { "type": "number" },
+      "recall_semantic_reserve": { "type": "number" }
     }
   },
   "activation": {
--- a/integrations/openclaw/plugin/src/index.ts
+++ b/integrations/openclaw/plugin/src/index.ts
@@ -42,6 +42,7 @@ const typeboxConfigSchema = Type.Object(
     expose_tools: Type.Optional(Type.Boolean()),
     recall_limit: Type.Optional(Type.Number()),
     recall_max_tokens: Type.Optional(Type.Number()),
+    recall_semantic_reserve: Type.Optional(Type.Number()),
   },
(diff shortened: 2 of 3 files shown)


joryirving:
Superseded by v0.4.14's `MEMINI_RECALL_SEMANTIC_RESERVE` — implementing the durable-tier reserve server-side is the cleaner approach (it applies before the limit, so no client-side over-fetch + recompose). Thanks for picking up the idea. Closing this plugin-side version.

PR #31, opened 2026-07-08
Make memory reliable and available to the LLM: MCP annotations, memory_update, degrade-not-fail writes, bounded recall

perfectra1n:
# Make memory reliable and available to the LLM

## Why

memini already had good bones: hybrid vector plus keyword recall fused with RRF, tiered memory with promotion and decay, async LLM consolidation, and a Claude Code plugin that injects context at session start. What it didn't have was a hardened path between the model and the memory. Three problems motivated this branch:

1. The MCP tools carried no annotations, so clients like Claude Code could not auto-approve read-only calls. Every recall interrupted the user with a permission prompt, which in practice means the model stops using memory mid-session.
2. Failure behavior was asymmetric. Recall degraded gracefully to keyword-only search when the embeddings endpoint went down, but writes failed outright. An embeddings outage silently stopped memory capture while recall kept limping along, and nothing told the model or the operator that anything was wrong.
3. The when-to-use guidance lived only in the Claude Code plugin skills. Any client connecting over bare MCP (Codex, Hermes, anything without the plugin) saw one-line tool descriptions and had to guess.

Before writing any code I surveyed what the ecosystem converged on: Anthropic's tool-writing guidance, the MCP 2025-06-18 spec, and the tool surfaces of mem0/OpenMemory, Zep/Graphiti, Letta, and LangMem. The design decisions below each trace back to one of those sources; the full list is at the bottom.

### MCP tool surface (`internal/api/mcp/mcp.go`)

**Annotations and titles on every tool.** `memory_recall`, `memory_get`, `memory_list`, `memory_briefing`, and `memory_answer` now declare `readOnlyHint: true`; `memory_forget` declares `destructiveHint: true`; `memory_remember` and the new `memory_update` are additive and idempotent. Everything sets `openWorldHint: false` (the MCP spec's own example of a closed-world tool is a memory tool). This is the single biggest availability win in the branch: clients that honor annotations stop prompting for every read.

**Tool descriptions rewritten as usage protocol.** Anthropic reported that refining tool descriptions alone moved Claude to state of the art on SWE-bench Verified, so the descriptions now carry the when-to-call, when-not-to-call, and what-comes-back guidance that previously existed only in the plugin skills. For example, recall's description now says "Empty results mean nothing is known: proceed from first principles, never invent a remembered fact," which is exactly the abstention behavior LongMemEval scores.

**Recall results carry `created_at` and `tags`.** The skills always told the model "on conflicting memories prefer the most recent," but recall items had no timestamp, so the instruction was impossible to follow. Additive fields, no breakage.

**`memory_list` is capped.** The old behavior was `limit: 0` returns every memory in the namespace, which for an LLM caller is a context blowout waiting to happen. Over MCP the default is now 20 with an `offset` parameter for paging (REST keeps 0 = all). A review pass caught that a negative offset could arithmetic its way past the cap and return everything; offsets are clamped to 0.

**New `memory_update` tool.** The merge-hint flow in the docs referred to a `memory_update` that never existed anywhere, not even in REST. The tool now exists over MCP as a partial update by id (only provided fields change, metadata merges key by key, `tags: []` clears while omitting keeps), composed of `Get` plus the existing `Remember`-with-id upsert path so it inherits re-embedding and the write-time lifecycle. mem0 and Graphiti both expose update tools; correcting a fact beats delete-and-rewrite because history and corroboration state survive.

**Degradation is visible in results.** When recall falls back to keyword-only search, the result now says so: `"degraded": "keyword_only"` plus a human-readable note telling the model the results may be incomplete. Same pattern on writes: a memory stored without a vector during an embeddings outage returns `"degraded": "pending_embed"`. This came out of research on memory-induced tool drift; a model acting on silently incomplete recall makes worse decisions than one that knows the search was partial.

**Errors are prompts.** Every error string now tells the model what to do next. `memory_get` with an unknown id says "ids come from memory_recall or memory_list results" instead of a bare not-found; invalid tier errors list the four valid tiers. The final review caught that `memory_update` wrapped every store error in the not-found wording, which would have told the model a memory doesn't exist during a transient store hiccup; the friendly wrap is now gated on `errors.Is(err, store.ErrNotFound)` with one shared helper across get, forget, and update.

**Optional `response_format: "concise"` on recall** returns the summary or the first 240 runes of each memory for cheap scanning, with `memory_get` as the drill-down for full text. This is the pattern Anthropic recommends and what claude-mem landed on after their v3 injected too much and polluted context. A review pass caught the initial implementation deciding truncation on byte length but cutting on rune length, which appended a false ellipsis to any multi-byte content; both now use rune length.

**`memory_answer` only registers when an LLM is configured.** Previously it was a guaranteed-error tool in headless deployments.

### Write-path reliability (`internal/service/`, `internal/store/`)

**Stores accept vectorless rows.** Both drivers rejected rows without an embedding (Postgres even declared `embedding vector(N) NOT NULL`). Now `len(embedding) == 0` stores the row with its keyword index entry and no vector entry. Postgres gets an idempotent `ALTER TABLE ... DROP NOT NULL` migration and an `embedding IS NOT NULL` filter in vector search. Self-review of this change found a real pre-existing seam: sqlite's `Reassign` used an INNER JOIN against the vector table and would have silently failed to move vectorless rows between namespaces; it's a LEFT JOIN now, with a regression test in the shared store conformance suite that runs against both drivers.

**Writes degrade instead of failing.** `Remember` wraps the content embed in `MEMINI_WRITE_EMBED_TIMEOUT` (default 5s). On timeout or error the memory is stored anyway, marked `pending_embed` in metadata, counted in the new `memini_remember_degraded_total{reason}` metric, and logged at Warn. Keyword search finds it immediately. Setting the knob to 0 restores the old fail-fast behavior. Write-time similarity jobs (dedup, corroboration, contradiction, sync consolidation) are guarded to skip vectorless rows; the implementation pass found that sync-mode consolidation also did a raw vector search and would have failed exactly the writes this feature rescues.

**A backfill loop heals the store.** `Service.RunEmbedBackfill` (interval `MEMINI_BACKFILL_INTERVAL`, default 1m, same ticker pattern as the existing promoter) finds pending rows across namespaces, re-embeds up to 100 per tick, strips the flag, and exports a `memini_embed_backfill_pending` gauge. If the first row of a tick fails to embed, the tick aborts on the assumption the embedder is still down, so a dead endpoint produces one warning per tick rather than a hundred. Two review findings hardened this: each row's embed is bounded by the write-embed timeout (a stalled but non-erroring embedder previously would have wedged the loop forever), and the row is re-fetched immediately before upsert with an `UpdatedAt` comparison so a concurrent `memory_update` landing during the up-to-5s embed window can't be clobbered by a stale snapshot. A healthy re-embed also deletes a stale `pending_embed` flag, so a recovered row can't keep reporting itself degraded.

### Read-path reliability

**Query rewrite is bounded and parallel.** The recent query-expansion feature (`query_rewrite`) called the LLM with no timeout of its own, so an enabled rewrite could hold a recall for the LLM client's full 120s HTTP timeout, and then ran each variant recall serially. The expansion call now gets `MEMINI_RECALL_REWRITE_TIMEOUT` (default 3s, falls back to plain single-query recall), and variant recalls run concurrently through an errgroup bounded by the existing recall concurrency limit, preserving variant order for the RRF fuse. One subtlety worth calling out for reviewers: the `Degraded` out-param added earlier in the branch is a shared pointer, and naively parallelizing would have raced on it. Each variant gets its own slot and the coordinator aggregates once after the join; there's a dedicated test that fails under `-race` if anyone regresses this.

**REST requests get a deadline.** `MEMINI_REQUEST_TIMEOUT` (default 60s, 0 disables) applies chi's timeout middleware to the `/v1` group only. `/mcp` is deliberately excluded because the Streamable HTTP transport holds a long-lived SSE stream, which is the same reason the server never sets a WriteTimeout. The default is 60s rather than 30s because `/v1/answer` at high reasoning effort legitimately rides the LLM's 120s budget; the tradeoff is documented in `config.go`. Known follow-up: a timed-out request currently surfaces as 500 rather than 504 because the error mapper doesn't special-case `context.DeadlineExceeded`.

### Operational visibility

**`/healthz?verbose=1`** reports per-dependency status: store (live ping), embedder and LLM (last success, last error, tracked by thin recording decorators wrapped around the real clients at startup). `/readyz` deliberately stays store-only; failing readiness on an embeddings outage would turn a degraded-but-working service into a total memory blackout, which is the opposite of what this branch is for. Because the verbose payload contains dependency error strings (internal hostnames, driver errors) on an endpoint that sits outside bearer auth by design, the verbose view requires the API key when one is configured and silently degrades to the plain liveness body otherwise, keeping probes working.

**Boot probe.** A bad `MEMINI_EMBED_BASE_URL` used to surface only on the first write. `buildServiceStack` now fires a one-shot embed at startup and logs a clear warning if it fails.

### Documentation and every other prompt surface

The plugin skills, README config table, and `docs/tiers.md` were synced with all of the above (this also flushed out two stale claims: the episodic TTL docs still said 90 days after the recent reduction to 30, and the recall skill referenced a result shape that hasn't existed for a while). Then a second pass swept every host integration (opencode, pi, Hermes, OpenClaw, Open WebUI, Codex) so their duplicated tool descriptions and injected prompts match reality. That sweep was careful about a subtle correctness rule: `memory_update`, `response_format`, and `offset` exist only over MCP, and REST has no update endpoint at all, so the REST-backed hosts explicitly do not advertise them and instead document forget-plus-re-remember for corrections. It also fixed four host descriptions that called forget a "soft delete (tombstone)" when `svc.Forget` is a hard delete.

## New configuration

| Env var | Default | What it does |
| --- | --- | --- |
| `MEMINI_WRITE_EMBED_TIMEOUT` | `5s` | Bounds the content embed on writes; past it the memory stores vectorless and pending. `0` = old fail-fast. |
| `MEMINI_BACKFILL_INTERVAL` | `1m` | How often the vector backfill loop runs. `0` disables. |
| `MEMINI_RECALL_REWRITE_TIMEOUT` | `3s` | Bounds the LLM query-expansion call on `query_rewrite` recalls. |
| `MEMINI_REQUEST_TIMEOUT` | `60s` | Request deadline on `/v1` routes only. `0` disables. |

## Compatibility

No breaking changes. Tool names are unchanged, all new result fields are additive and omitted when empty, REST keeps its existing semantics (including `limit: 0` = all on list), the briefing arg change (`int` to `*int` so an explicit 0 can disable a section over MCP) keeps the same wire shape, and the Postgres migration is an idempotent `DROP NOT NULL` that is a no-op on fresh databases.

## How it was verified

Every task was built test-first, and each diff went through an adversarial review before landing. That process caught ten real bugs before merge, including four that only existed for the length of one review cycle: a negative list offset that bypassed the new cap, a nil-pointer dereference in `memory_update` that would crash the whole stdio MCP server when the episodic value gate dropped a rewrite, the byte-versus-rune truncation bug in concise mode, and the missing embed timeout that could wedge the backfill loop behind a stalled endpoint.

Final state: the full CI battery was reproduced locally at the head commit with the same VectorChord Postgres image CI uses (`ghcr.io/tensorchord/vchord-postgres:pg18-v1.1.1`), and everything passes: go.mod tidiness, generated REST code drift, the full race-enabled suite, the Postgres conformance suite (the `DROP NOT NULL` migration, NULL vector insert, and vector-search NULL filter all behave correctly against a live database), the end-to-end integration tests that boot the whole server stack, the bench harness compile check, and the plugin script tests. The five host integration test suites all pass as well (opencode 28, pi 12, OpenClaw 53 and 41, Hermes 32, Open WebUI 15).

## Known follow-ups (tracked, not blocking)

- Backfill poison-row starvation: a row whose content permanently fails to embed leads every tick and blocks the backlog behind it. Distinguishing transport errors from content errors, or a per-row retry cap, would fix it.
- `/v1` timeouts return 500 instead of 504.
- `memory_answer` results carry no degraded marker even when their internal recall was keyword-only.
- The embedder health block has no `configured` field, and client-cancelled requests count as dependency failures in the tracker, which can make verbose healthz flap.
- A/B the parallel rewrite path with `cmd/bench -rewrite` to quantify the latency win.

## Sources

- Anthropic, "Writing effective tools for agents" (descriptions as the highest-leverage knob, token-efficient results, response_format pattern, errors as prompts): https://www.anthropic.com/engineering/writing-tools-for-agents
- MCP specification 2025-06-18, Tools (annotations; the closed-world memory-tool example): https://modelcontextprotocol.io/specification/2025-06-18/server/tools
- Anthropic, "Effective context engineering for AI agents": https://www.anthropic.com/engineering/effective-context-engineering-for-ai-agents
- mem0 / OpenMemory MCP tool surface (update_memory precedent): https://docs.mem0.ai/platform/mem0-mcp
- Zep/Graphiti MCP server (health visible to the agent, bi-temporal invalidation): https://github.com/getzep/graphiti/blob/main/mcp_server/README.md
- Letta on agent memory (always-present summary plus on-demand search): https://www.letta.com/blog/agent-memory/
- claude-mem hooks architecture and the SuperBrain writeup (context pollution from over-injection; drill-down pattern): https://docs.claude-mem.ai/hooks-architecture and https://alexandrekhoury.com/writing/superbrain-session-memory-claude-code
- LongMemEval (abstention as a scored ability): https://arxiv.org/pdf/2410.10813
- MemFail, stress-testing memory systems (pollution, staleness, over-remembering): https://arxiv.org/pdf/2605.26667
- Memory-induced tool drift (why degraded recall must be visible): https://arxiv.org/pdf/2605.24941
- modelcontextprotocol/go-sdk v1.6.1 source, for the exact annotation field shapes.


eleboucher:
> _Originally by **eleboucher** on 2026-07-08 19:34 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/82#issuecomment-11438)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #83), which is the source of truth. Further review and merging happen there.

PR #33, opened 2026-07-10
fix(plugin): stop leaking <memory> blocks into replies, save via memory_remember

perfectra1n:
## Problem

The SessionStart directive currently asks the agent to print a raw `<memory>{"memories":[...]}` block at the end of its reply so the Stop hook can scrape it out of the transcript. The capture works, but the XML+JSON shows up in the terminal every time the agent saves something. I kept noticing it in my own sessions and it looks like broken output, not a feature.

## What changed

The directive now points the agent at the `memory_remember` MCP tool instead (tier "semantic", one memory per call) and tells it to never print memory markup in the reply. Tool calls render as a collapsed line in Claude Code, so saves no longer show up in the reply text. The wording follows the auto-save nudge, which already asks for `memory_remember`.

The transcript scraper in the Stop hook stays as-is, as a fallback: sessions that started under the old directive (or a model that emits a block anyway) still get captured with the same `inline_extract` provenance. `MEMINI_INLINE_EXTRACT` keeps gating both the injection and the scrape, so setting it to `0` still turns off model-curated capture entirely.

One detail worth calling out: the new instruction contains no literal `<memory>` tag anywhere, so there is nothing for a model to echo back. A new test pins that, along with the tool names and tier guidance, and the session-start test now checks the injected context carries the directive without markup. I also renamed the wrapper tag from `<memini-inline-memory>` to `<memini-memory-directive>`, which makes it easy to spot old vs new sessions in transcripts.

If the MCP tools are not available the directive says to do nothing. Turn capture, checkpoints, and distill-on-write are untouched and still record the session.

## Docs

Updated the hook table, the automatic-capture bullet, and the env var row in `plugin/README.md` for the tool-based flow, and added a note about pre-approving `mcp__plugin_memini_memini__memory_remember` in `permissions.allow` so the first save does not stall on a permission prompt.

## Tests

`node --test plugin/scripts/_test.mjs`: 51 pass, 0 fail.

### DIFF
--- a/plugin/README.md
+++ b/plugin/README.md
@@ -12,7 +12,7 @@ the agent _when_ to use the memory tools.
 | `SessionStart` | Searches prior context, writes a short block to the agent's input                                                                                                             |
 | `PreToolUse`   | Before Edit/Write/Read/Glob/Grep, surfaces related memories                                                                                                                   |
 | `PostToolUse`  | Buffers state-changing tool calls locally (no network, no per-call memory)                                                                                                    |
-| `Stop`         | Distills the buffer into a working-tier checkpoint, captures the last turn as episodic memory, extracts inline `<memory>` facts, and periodically nudges an auto-save (below) |
+| `Stop`         | Distills the buffer into a working-tier checkpoint, captures the last turn as episodic memory, scrapes legacy inline `<memory>` blocks (back-compat), and periodically nudges an auto-save (below) |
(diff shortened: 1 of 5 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-07-10 05:33 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/85#issuecomment-11637)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #86), which is the source of truth. Further review and merging happen there.

PR #35, opened 2026-07-10
Teach agents the conventions memini already has: server instructions, schema enums, and correction-by-upsert

perfectra1n:
I spent some time auditing every string memini shows to a model (tool descriptions, parameter schemas, the plugin's injected directives, the integration plugins) against Anthropic's tool-writing and skill-authoring guidance and the MCP spec, and then checked the theory against a real deployment: my production memini instance, running 0.6.4, with 182 memories accumulated across 8 namespaces. The good news is that the existing descriptions hold up well. The trigger language ("call BEFORE starting work that may have history"), the degraded-mode caveats, and the read-only framing on injected blocks are all things the guidance recommends and memini already does. But the live data surfaced a handful of gaps where the prompts and reality disagree, and they were all measurable in the stored data.

## What the deployment data showed

**The `pinned` tag was used once in 182 memories, and `metadata.category` almost never.** Every session briefing was rendering with an empty Pinned section. The interesting part is *why*: both conventions are taught only in the optional `remember` skill, which a model frequently never loads before calling `memory_remember` directly. The one namespace where categories were set consistently (20 of 24 memories) was the one where I had explicitly asked the agent to tag things properly in the prompt. So models follow these conventions fine when told; the tool schema just never tells them.

**The procedural tier was starving: 7 procedural vs 89 semantic.** Clearly procedural content (toolchain workarounds, command recipes) was sitting in the semantic tier. This traces directly to the plugin's session-start directive and auto-save nudge, which both hard-coded `tier "semantic"` for every save.

**7 of 31 auto-captured turns in my main namespace were junk.** `isRealUserMessage` predates background-task notifications, so `<task-notification>` and `[SYSTEM NOTIFICATION - NOT USER INPUT]` blocks from the harness were being captured as episodic "user" turns, complete with task ids and output-file paths, and then surfacing in recall results and the briefing's recent-activity section.

## What changed and why

**The MCP server now sends `instructions` in the initialize response** (`NewServer` was passing nil options). Per-tool descriptions are read in isolation during tool selection, so they can't express cross-tool policy: call `memory_briefing` at session start, recall before work that may have history, remember proactively, correct with `memory_update` instead of writing near-duplicates. Server instructions are the one place that policy can live for a bare MCP client that doesn't have the plugin's hooks and skills. This is also where `pinned` and `metadata.category` are now taught, alongside the parameter descriptions themselves (`tags` and `metadata` now carry the conventions in the schema, where every client sees them).

**Constrained params now carry real JSON Schema enums.** `tier`, `level`, `scope`, `response_format`, `reasoning_level`, and the `tiers`/`levels` filter items previously documented their legal values in prose ("any other value is rejected as an error"). jsonschema struct tags can only hold a description, so the enums are applied by building the inferred schema explicitly and setting them via `Tool.InputSchema` before `AddTool` (see `enumSchema` in mcp.go). One wrinkle worth knowing about: for an `omitempty` slice the enum has to go on `Items`, and the check has to be `Items != nil` rather than `Type == "array"`, because the inferred schema uses the multi-type `Types` field there. Getting that wrong makes every filtered call fail schema validation, which the new tests would catch. A side effect of this change is that invalid enum values are now rejected by the SDK's schema validation before reaching the handler, so the error text for a bad tier differs from the old server-side message. The server-side validation stays as a backstop.

**Recall's description drops the `exclude_metadata` session-id advice.** It told the model to pass its own session id to avoid seeing its captured turns echoed back, but a model has no way to know its session id, the plugin's hooks already exclude it client side, and the server has a temporal echo guard on top. It was dead weight in every session's token budget. `exclude_metadata` keeps a usable example instead.

**The plugin directives are tier-aware and the capture filter is stricter.** The session-start directive and auto-save nudge now route facts/decisions/preferences to semantic and how-tos to procedural, teach `pinned`, and tell the model to skip anything already covered by CLAUDE.md or project docs. `isRealUserMessage` rejects the two harness-notification shapes above, with test cases for both.

**The REST integrations stop recommending forget-and-re-remember for corrections.** openclaw, pi, hermes, and openwebui all claimed the only way to correct a fact over REST is to delete it and write a replacement, because there is no partial-update endpoint. That's technically true but stale in effect: `POST /v1/memories` upserts when given an `id`, which keeps the memory's identity and history where forget+re-remember destroys both. All four now accept an optional `id` on `memory_remember`, their forget descriptions point there, and `integrations/README.md` explains the actual distinction (the MCP `memory_update` merges field by field, the REST upsert replaces what you send). While in there I also ported the stronger trigger language these descriptions had fallen behind on ("empty results mean nothing is known, never invent a remembered fact", the don't-store-what's-in-docs caveat, and `pinned`).

One drive-by fix: the openclaw `helpers.test.ts` mock was missing `deleteJson`, so `pnpm typecheck` in that package was already failing on main before this branch. The mock now implements it.

## Verification

- `go build ./...` and `go test ./...` pass (new tests pin the instructions on initialize and the enums on tools/list, alongside the existing description-phrase pins)
- `golangci-lint` v2.12.2 (the CI version) reports 0 issues
- `node --test plugin/scripts/_test.mjs`: 60 pass, including the two new `isRealUserMessage` cases
- `pnpm -r build`, `pnpm -r typecheck`, `pnpm -r test`: all green (openclaw 53+41, pi 15, namespace-resolver 33)
- hermes `python3 -m unittest`: 38 pass
- stdio smoke test against the built binary: initialize carries the instructions, tools/list shows the enums and the new parameter descriptions on the wire

Happy to split this into separate PRs if you'd rather take the server changes and the integration changes independently; the three commits are already cut along those lines.

### DIFF
--- a/integrations/README.md
+++ b/integrations/README.md
@@ -117,13 +117,15 @@ is the exception: its native plugin is deliberately tool-free (automatic recall 
 capture only), so to give it `memory_forget` (or any tool) wire the memini MCP
 server alongside the plugin.
 
-**`memory_update` (partial update by id) is MCP-only** — there is no REST
-update endpoint. Claude Code, Codex, and any host wired to the memini MCP
-server get it; Hermes, Pi, OpenClaw, Open WebUI, and opencode's native plugins
-talk REST (`/v1/search`, `/v1/memories`) and do **not** expose it, so their
-tool descriptions correctly point to `memory_forget` (delete + re-remember) as
-the only way to correct a stored fact. Wire the MCP server alongside a native
-plugin if a host needs `memory_update`.
+**`memory_update` (partial update by id) is MCP-only** — REST has no PATCH
+endpoint, but `POST /v1/memories` **upserts when given an `id`**, replacing the
+memory in place while preserving its identity and history. The REST-backed
+plugins (Hermes, Pi, OpenClaw, Open WebUI) therefore accept an optional `id` on
+`memory_remember`, and their tool descriptions point to re-remember-with-id as
+the way to correct a stored fact; `memory_forget` is only for memories that
+should not exist at all. Wire the MCP server alongside a native plugin if a
+host needs true partial updates (`memory_update` merges field-by-field; the
+REST upsert replaces the record with what you send).
 
 Two host-specific differences remain by design:
 
--- a/integrations/hermes/plugin/memini/__init__.py
+++ b/integrations/hermes/plugin/memini/__init__.py
@@ -610,10 +610,11 @@ class MeminiMemoryProvider(MemoryProvider):
                 "name": "memory_recall",
                 "description": "Search long-term memory (memini) for relevant past facts and context. Call "
                 "before starting work that may have history: editing an unfamiliar file, "
-                "debugging a recurring issue, or when asked what's known about something. A "
-                'degraded:"keyword_only" field in the result means semantic search was '
-                "unavailable and results came from keyword matching alone — treat as "
-                "incomplete, not exhaustive.",
+                "debugging a recurring issue, or when asked what's known about something. "
+                "Empty results mean nothing is known — proceed from first principles, never "
+                'invent a remembered fact. A degraded:"keyword_only" field in the result '
+                "means semantic search was unavailable and results came from keyword "
+                "matching alone — treat as incomplete, not exhaustive.",
                 "parameters": {
                     "type": "object",
                     "properties": {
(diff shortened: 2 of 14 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-07-10 21:16 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/88#issuecomment-11813)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #90), which is the source of truth. Further review and merging happen there.

PR #36, opened 2026-07-11
Scope redesign: ancestor cascade, home namespaces, and explicit links

perfectra1n:
## Scope redesign: ancestor cascade, home namespaces, and explicit links

This PR replaces the four scope mechanisms that grew up around namespaces (`MEMINI_GLOBAL_NAMESPACE`, `MEMINI_TENANT_SHARED` with the `_shared` naming convention, `scope=subtree`, and per-call namespace lists) with one model: writes land in exactly one namespace, and reads resolve through an always-on cascade. It is a breaking change, and I think pre-1.0 is the right time for it. There is a `memini migrate scopes` command and a boot-time refusal that walks operators through the transition.

### Why the old model was hurting

I have been running a shared memini instance for a while now: my own project namespaces alongside a team tree that several people write into. Four problems kept coming up, and they all trace back to the same root cause, which is that the scope semantics were bolted on one special case at a time.

**Tenancy stopped at depth 1.** Namespaces themselves are slash paths of any depth, but `tenantSharedNamespace` cut at the first segment only. Once a real tree grew to three levels (think `acme/phoenix/api` under `acme/phoenix` under `acme`), the middle level could not share anything downward. The syntax promised a tree, the semantics delivered a two-level list.

**Users invented the cascade themselves, in the wrong place.** Watching actual usage, nobody wrote shared facts into `<tenant>/_shared`. People wrote org-wide facts directly into the tenant root namespace, because that is where they intuitively belong. The `_shared` sibling convention fought how the tool actually gets used.

**Shared instances have no person axis.** A single global namespace is per-instance, but "always remember this about me" is per-person. On an instance where two people work in the same team tree, one person's global preferences would leak into the other's briefings. Meanwhile, per-person session digests (episodic memories) were landing in shared namespaces and polluting them.

**Agents could not reason about placement.** The namespace was invisible plumbing set by headers. The clearest signal that this had failed: the shared namespace on my instance ended up containing a stored procedural memory whose content was instructions teaching agents where to write. When the data layer has to document the API, the API is wrong.

### The new model

One rule replaces everything. A write goes to exactly one namespace (unchanged, this invariant is the best thing about the original design). A read in namespace N sees:

1. N itself, all tiers
2. every ancestor of N, nearest first, durable tiers only (`acme/phoenix/api` reads `acme/phoenix`, then `acme`)
3. the caller's home namespace (new `X-Memini-Home` header / `MEMINI_HOME` client env), durable only
4. explicit namespace links, one hop, durable by default with optional per-link tier restriction

The interior nodes of the tree ARE the shared layer, which matches how people already use it. The global namespace becomes per-person (home). Lateral, non-tree sharing gets a first-class primitive (links) instead of env var contortions.

```mermaid
flowchart LR
    subgraph read ["recall from acme/phoenix/api"]
        P["acme/phoenix/api\nprimary, all tiers"]
        A1["acme/phoenix\nancestor, durable"]
        A2["acme\nancestor, durable"]
        H["personal/kit\nhome, durable"]
        L["shared/golang\nlink, durable"]
    end
    P --> F["fusion + rerank\nnearest wins ties"]
    A1 --> F
    A2 --> F
    H --> F
    L --> F
    F --> R["results with provenance:\n(from: acme), (from: personal),\n(from: link:shared/golang)"]
```

On the write side, the agent never constructs a namespace path. `memory_remember` takes `visibility`: `project` (default), `personal` (goes to your home namespace), or an ancestor's name. Episodic and working memories are clamped to the project namespace no matter what visibility says, because a session digest is inherently first-person and situational. That clamp is what stops per-person noise from polluting shared levels.

```mermaid
flowchart TD
    W["memory_remember\nvisibility: 'acme'"] --> C{"final tier durable?"}
    C -- "no (episodic/working)" --> P2["clamped to primary\nacme/phoenix/api"]
    C -- yes --> V{"visibility"}
    V -- "project / empty" --> P3["acme/phoenix/api"]
    V -- personal --> HM["personal/kit\n(errors if MEMINI_HOME unset)"]
    V -- "ancestor name" --> AN{"matches chain?"}
    AN -- yes --> T["acme"]
    AN -- no --> E["error listing valid targets:\nproject, personal, acme/phoenix, acme"]
```

That error message is deliberate. The valid-targets enumeration is how the model learns the topology, so a wrong guess becomes a correction instead of a dead end. The briefing does the same job from the other side: it now opens with a scope line like `Scope: acme/phoenix/api ← acme/phoenix(3) ← acme(4) ← personal(2), +1 link`, so after one briefing the agent has seen its whole scope world.

### Why these particular shapes

A few decisions worth explaining, since alternatives were on the table:

**Paths stay, dimensions lost.** I considered replacing paths with typed scope fields (user/agent/project, the mem0 style). Storage-wise paths won: the vec0 partition key and the FTS layout are already path-native, and real data is genuinely hierarchical. But the LLM-facing surface borrows the dimension idea, since a model choosing from `project | full | everywhere` is dependable and a model assembling path lists is not. Paths are the physics, semantic enums are the steering wheel.

**Links are one hop, permanently.** If A links to B and B links to C, A does not see C. Every transitive design I looked at makes "why did recall return this?" unanswerable in one lookup. Provenance stays explainable; that felt worth more than expressive power.

**Home is client-asserted, not enforced.** This stays scoping, not authz. One API key per instance is unchanged. The model is designed so a permission boundary could later attach at a subtree root, but nothing here pretends to be access control.

**REST keeps raw namespaces; only the MCP tools lose them as choices.** Scripts and the admin UI legitimately address namespaces. The distinction the code draws is choosing versus addressing: result items still carry their namespace so `memory_update`/`memory_forget` can copy it verbatim, but no LLM-facing tool takes a path as an input choice anymore. REST also keeps `exact`/`subtree` as deprecated aliases so existing callers do not break.

### Who owns what

```mermaid
flowchart TD
    subgraph client ["client (plugins, integrations)"]
        NS["resolve namespace\n(env, git remote, cwd)"]
        HO["resolve MEMINI_HOME"]
    end
    subgraph server ["server (service layer)"]
        RS["cascade + link + call\nread-set resolution"]
        VIS["visibility resolution\n+ tier clamp"]
        PROV["provenance labels"]
    end
    subgraph store ["store (sqlite-vec / Postgres)"]
        PART["namespace partitioning"]
        LNK["namespace_links table"]
        AGG["per-namespace activity\naggregate"]
    end
    NS -- "X-Memini-Namespace" --> RS
    HO -- "X-Memini-Home" --> RS
    RS --> PART
    RS --> LNK
    VIS --> PART
    RS --> PROV
```

Clients only resolve identity and send headers. The server owns every scope decision. The store partitions and knows nothing about cascades. Server-side scope configuration is now zero env vars.

### Breaking changes and migration

* `MEMINI_GLOBAL_NAMESPACE` and `MEMINI_TENANT_SHARED` are gone. The server (and `memini mcp`) refuse to boot if either is set, with a message pointing at the migration path, so nobody silently loses their old sharing semantics.
* `memini migrate scopes` folds every `<tenant>/_shared` into `<tenant>` (dry-run by default, `--yes` to apply, idempotent, and it runs a content dedup pass on each merge target since moving by ID alone would duplicate restated facts).
* The old global namespace is adopted rather than rewritten: single-operator instances set `MEMINI_HOME=<old value>` on their clients, team-wide ones use `memini link add`.
* The widening to be aware of: nested namespaces now always read their ancestors (durable tiers only). The old default was isolation unless you opted in. Worth leading with in the release notes for whatever version ships this.

### Surface changes

New: `GET /v1/namespaces/read-set` (introspection with origins), `/v1/links` CRUD, `memini link add|rm|ls`, `memini migrate scopes`, `visibility` on remember, `scope` on search/briefing/answer (MCP), `from` provenance on results, briefing scope header and child rollup, doctor rewritten for the new resolution (it prefers the server endpoint and warns about the catch-all `MEMINI_NAMESPACE` pin and unset `MEMINI_HOME`). The admin UI grew a real n-level namespace tree (deep namespaces used to render flat under the first segment), a Scopes view, provenance chips, and a namespace mode for the graph. `docs/scopes.md` documents the whole model with worked examples.

### Testing

Every task on this branch was test-first with the failing run captured before implementation. The store changes have conformance coverage that runs against both backends, and the Postgres side was exercised against a real VectorChord container, not just sqlite. Beyond unit coverage I ran a full end-to-end pass against a scratch server: seeded cascade, read-set ordering, visibility routing and clamp, migration dry-run/apply/re-run, boot refusals, doctor output, export/import round-trip including links. The retrieval benchmark (offline deterministic sample) is byte-identical before and after, which was the main thing I wanted to prove since cascade legs change fusion candidate pools. Plugin and integration suites are green across the board; fixing the plugin test isolation (ambient `MEMINI_*` env leaking into spawned hooks) is its own commit early in the branch.

### Known follow-ups

Two capability gaps are documented rather than fixed here, both called out in `docs/scopes.md`: the agentic answer loop's `keyword_search` tool still searches only the primary namespace (its sibling semantic search uses the full cascade), and REST `/v1/answer` does not yet expose `scope` (MCP does; REST always uses the default full cascade).

I split the work into small commits (store, service, transport, MCP, CLI, docs, UI) so it should review in layers. The two files that carry the design are `internal/service/readset.go` and `internal/service/visibility.go`; everything else hangs off them. Happy to split anything out or rework any decision.


eleboucher:
> _Originally by **eleboucher** on 2026-07-11 16:28 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/91#issuecomment-11910)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #92), which is the source of truth. Further review and merging happen there.

PR #37, opened 2026-07-11
Named API keys with per-key home and default namespace

perfectra1n:
## Named API keys with per-key home and default namespace

This builds on the scope redesign from #36. That PR gave every person a home namespace, but identity was still client-asserted: whoever holds the shared bearer token can claim any home by setting `MEMINI_HOME`. On a shared instance that means one typo in someone's shell profile quietly reads and writes another person's personal memories. This PR makes identity server-side: an instance can now have multiple named API keys, and each key can carry its own home namespace and default namespace.

### What a key is

A key is identity, not authorization. Any valid credential can still read and write any namespace, exactly like the single shared token always could. What changes is that the server now knows who is calling, which buys four things:

* per-person homes with zero client config (no `MEMINI_HOME` needed; the binding lives on the key)
* no accidental cross-person home claims (a bound key's home wins and the `X-Memini-Home` header is ignored, logged if it disagrees)
* revocation and rotation per person instead of rotating one shared secret for everyone
* attribution: writes from a named key stamp the key's name into `metadata.author` unless the caller set one, so "who recorded this decision" is answerable on shared namespaces

There are three kinds of credential. The existing `MEMINI_API_KEY` env var becomes the admin/bootstrap key and behaves exactly as before. Named keys live in a new `api_keys` table, managed by CLI, REST, or the UI. Declarative keys come from a YAML file (`MEMINI_API_KEYS_FILE`) loaded at boot, which suits GitOps deployments where the file is a SOPS-encrypted secret; file keys shadow same-named table keys and are immutable through the API.

```mermaid
flowchart TD
    B["bearer token"] --> A{"admin key?\n(constant time)"}
    A -- yes --> AD["admin: header-driven,\nmay manage keys"]
    A -- no --> F{"file key?\n(SHA-256 lookup)"}
    F -- yes --> P["principal: name,\nhome, default ns"]
    F -- no --> T{"table key?"}
    T -- yes --> P
    T -- no --> R["401"]
    P --> H["home: key binding wins,\nheader ignored"]
    P --> N["namespace: header wins,\nkey default fills absence"]
```

That last asymmetry is deliberate. Home is identity, so the server-known binding beats anything the client says. The default namespace is context, so an explicit `X-Memini-Namespace` header still wins and the key's default only fills in when the header is absent. Both rules are documented where they live, in the middleware.

### Secrets

A secret is 32 bytes from `crypto/rand`, shown exactly once (CLI stdout, or the create/rotate response, or a one-time modal in the UI) and never stored; only a hex SHA-256 lands in the database. Generation and hashing each have one canonical helper in `internal/apiauth` so the CLI, REST, and file loader can never drift. Rotating a key (`memini key add <name>` on an existing name) regenerates the secret and preserves everything else, including disabled state, so rotating a key you disabled during an incident never silently re-enables it. Explicit flags still override.

### Key management and bootstrap

`/v1/keys` (and the UI's Keys view) is admin-gated: the admin key or unauthenticated dev mode can manage keys, named keys get 403. Dev mode enables a UI-first bootstrap: on a fresh open instance you create your first key in the browser, and auth turns on immediately, on both the REST and MCP surfaces (they share one auth resolver and cache, and mutations invalidate it). The fail-closed corollary is documented: if you bootstrap without ever setting an admin key, API key management locks itself out and the `memini key` CLI against the store is the escape hatch.

Everything else follows conventions this codebase already has: the store capability is an optional interface like `LinkStore`, `maintenance.Move` renames key bindings the way it already renames link endpoints, dangling bindings after a namespace deletion are legal (namespaces exist implicitly) and `memini doctor` notes them, and `docs/api-keys.md` documents the model with worked examples.

### Testing

Test-first throughout, with conformance coverage on both backends and the Postgres side run against a real VectorChord container. The auth work got an explicit security review pass per task: constant-time admin comparison, hash-before-lookup, no secret in any log or list response, fail-closed store errors, per-request principal (no cross-session leakage on the MCP streaming path), and an end-to-end bootstrap test proving auth enforcement is immediate after the first key is created, on both surfaces.

### Known follow-ups

Three small ones, all noted in code comments: the UI detects the admin-required state by matching the 403 body text rather than the status code; the MCP surface still TrimSpace-only normalizes `X-Memini-Namespace` while REST fully normalizes it (home headers are consistent on both, namespace is the pre-existing holdout); and the `api_keys` add-column migrations, verified by hand against real old-schema databases, deserve a committed regression test.

Happy to split or rework anything. The two files that carry the design are `internal/apiauth/apiauth.go` and `internal/apiauth/filekeys.go`.


eleboucher:
> _Originally by **eleboucher** on 2026-07-11 20:51 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/94#issuecomment-11944)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #95), which is the source of truth. Further review and merging happen there.

PR #38, opened 2026-07-11
Activity log: record what memory was served and why, plus browse/activity filters

perfectra1n:
## Record what memory was served, and why

Ask memini "what did you use recently, and why?" and today it can't really answer. Usage is two counters on the memory row, `access_count` and `last_accessed_at`, bumped by `Reinforce` on the recall path. That tells you a memory was used. It doesn't tell you which question it answered, where it ranked, or how well it scored, and all of that exists at serve time: `Recall` computes fused and reranked scores, hands them to the caller, and throws them away. `store.Scored` carries a single float, and by the time results reach `reinforceResults` only the IDs survive.

The gaps are worse than "lossy". `memory_get` and `memory_briefing` don't reinforce at all, so `last_accessed_at` doesn't mean "last used", it means "last recalled". A memory served in every session briefing for a month looks untouched.

This PR adds a real activity log, an Activity page to read it, and (since you can't explore a feed you can't filter) filters on both that page and the browser.

### The log

A new `memory_events` table in both drivers, behind an optional `EventLogStore` capability interface. That follows the `LinkStore` / `APIKeyStore` precedent: the service and REST layer type-assert and degrade, so a driver without it just skips logging and `/v1/activity` answers 501.

One row per (operation, memory). Rows from one logical operation share an `op_id` and are inserted as a single batch, which is the load-bearing detail. A batch shares a `created_at` and gets contiguous ids, so `ORDER BY created_at DESC, id DESC` keeps an operation's rows adjacent. That adjacency is what lets the reader regroup a flat page of rows back into whole events by walking consecutive rows, with no join table and no second query.

One consequence worth knowing: within a group the rows come back in reverse rank order, because the id tiebreak runs backwards. `groupEvents` re-sorts by rank. There's a conformance case pinning it, and it caught me the first time I wrote the test.

Each row denormalizes a snapshot of the memory (namespace, tier, summary). That's deliberate on two counts. The feed renders in one query with no N+1 fetch per row, and a `forget` event stays readable after the memory it describes no longer exists. A recall that matched nothing gets a sentinel row too, because "this query found nothing" is exactly the sort of thing you open an activity feed to discover.

Writes are async and best-effort, off the request path, using the same `context.WithoutCancel` plus bounded-timeout shape as `reinforceResults`. A wedged log can slow nothing and fail nothing. `MEMINI_ACTIVITY_LOG=false` turns it off, and `MEMINI_ACTIVITY_RETENTION` (30d) and `MEMINI_ACTIVITY_MAX_ROWS` (100k) bound it, pruned by the existing decay sweeper.

### Logging is not reinforcing

This is the one design decision I'd argue about if someone proposed the alternative. It would have been easy to make briefings and gets bump `access_count` while we're in there, since they are reads after all.

They must not. `Reinforce` isn't a usage counter, it's a relevance signal. It slides TTLs forward and gates episodic to durable promotion (`promoteMinAccess`) and durable ranking. A briefing fires on every session start over the same top-N whether or not anything in it was relevant. Reinforcing that would inflate `access_count` uniformly across exactly the memories that already rank highest, and quietly corrupt promotion. `memory_get` is addressing a memory by ID, often the admin UI's own drawer, which says nothing about relevance either.

So both are logged but never reinforced. The activity log answers "when and how was this used" without touching the signal that decides what gets promoted. `TestBriefingAndGetLogButDoNotReinforce` pins it, because this is the kind of invariant someone helpful "fixes" in six months.

### The Activity page

Newest-first feed, grouped by operation, at `/activity`. A recall renders as its query plus the memories it served, each with rank and score, which is the "why". A briefing shows which section each memory came from. Forget and supersede show what was removed. Kind chips hide the noisy kinds (briefings dominate otherwise) and the selection persists, because re-hiding them on every visit is its own small misery.

Against a live instance it reads like this:

```
recall   "which database do we use"   served 1 memory     just now
           #1  semantic  the production database is postgres 16   1.00
recall   "zzz nonexistent kangaroo"   found nothing       1m ago
briefing  session briefing · 2 memories                   1m ago
           #1  semantic  the production database is postgres 16   facts
```

### Filters, and one honest bit of SQL

Browse had tier/tag/metadata filters and no sorting anywhere in the stack. Neither driver's `List` had an `ORDER BY`, so a single-namespace listing came back in insertion order while the "All projects" view sorted by `CreatedAt` in Go. There's now a real `sort` / `order` param (created, updated, last-used, recall-count, importance) mapped through a whitelist switch to an `ORDER BY` in both drivers, with an id tiebreak so a capped listing is deterministic, and the all-namespaces merge uses a Go comparator kept in lockstep with the SQL. Default is `created_at desc`, which is both sane and a bug fix.

Browse also gains memory-type (repeatable, OR semantics, because `meta=k=v` ANDs one value per key and can't express a multi-select), level (the REST layer already accepted it, the UI just never sent it), a recency window, and namespace narrowing in All-projects mode. The Activity feed gets kind, tier, free text, time window, and namespace.

Everything is server-side, which matters more than it sounds. The listing is capped at 500 under a sort, so a client-side recency filter would be correct only when sorting by that same time field and would quietly under-report otherwise. Same for narrowing an aggregate: filtering the response can starve a namespace whose rows fell outside the global cap.

The interesting case is tier and text on the activity feed. An event is a group of rows, so filtering rows would make a recall that served five memories render as "served 2 memories" under `tier=semantic`, which is a lie about what happened. Those two filters select whole operations instead, via `op_id IN (SELECT op_id FROM memory_events WHERE ...)`, and return the matching event intact. A conformance case and a REST case both assert that a tier-filtered recall still reports every memory it served.

### The home-header override now says something

#95 landed per-key home namespaces, where a bound key's home wins over a conflicting `X-Memini-Home` header and the disagreement is "logged if it disagrees". In practice that log was at debug level, so the override was invisible: you set a home, got a different one, and nothing told you why.

It now logs at warn and returns an `X-Memini-Warning` response header, which the admin UI renders as a banner:

> `home namespace "someone/elses/home" from X-Memini-Home ignored: API key "bound-bot" is bound to home "acme/home", which wins`

Still never a 400, since a shared client config that always sets the header is not a caller error, and no warning at all when the header agrees with the binding or the key is unbound. A warning that fires when nothing was overridden is noise, and noise gets ignored.

### Verification

Store conformance covers the new capability and filters for both drivers: sort ordering per key, limit-takes-the-top-N, memory-type OR, recency bounds, event round trip, op-adjacency, keyset cursor, prune by age and by cap, and that LIKE wildcards in a text filter stay literal instead of matching everything. Service and REST tests cover the log-versus-reinforce split, remember versus update, the forget snapshot outliving its memory, pagination, 400s on unknown sort/kind/tier values, and the home warning firing only on a real conflict.

Beyond the suite I drove a real server end to end (writes, recalls, briefings, forget, every filter, pagination) and confirmed the pages render and the filters actually reach the server rather than being applied in the browser.

### Notes for review

- No MCP tool for the activity log here. REST is what the UI needs, and a `memory_activity` tool is a clean follow-up if it's wanted.
- Opening a memory from the Activity page issues a `get`, which logs a `get` event. Self-noise, hidden by the kind chips. An opt-out header would fix it properly if it grates.
- Postgres shares the conformance suite but I ran the sqlite driver locally. The pg suite needs Docker (`mise run test-integration`).

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -137,9 +137,42 @@ paths:
           description: >
             Aggregate across every namespace, ignoring the namespace header. The
             server merges all namespaces and applies limit as a single global cap
-            (newest first), so the admin UI's "All projects" view fetches one
-            response instead of one request per namespace.
+            under the requested sort, so the admin UI's "All projects" view fetches
+            one response instead of one request per namespace.
           schema: { type: boolean, default: false }
+        - name: namespace
+          in: query
+          description: >
+            With all_namespaces=true, restrict the aggregate to these namespaces
+            (repeatable, exact match); ignored otherwise. Lets the browser narrow
+            an "All projects" listing without changing the active namespace.
+          schema: { type: array, items: { type: string } }
+        - name: memory_type
+          in: query
+          description: >
+            Repeatable and/or comma-separated metadata.memory_type filter; a memory
+            matches if its type is ANY of the listed values (OR). Distinct from
+            "meta", which ANDs one value per key.
+          schema: { type: array, items: { type: string } }
+        - name: created_after
+          in: query
+          description: Only memories created at or after this instant.
+          schema: { type: string, format: date-time }
+        - name: accessed_after
+          in: query
+          description: Only memories last accessed at or after this instant.
+          schema: { type: string, format: date-time }
+        - name: sort
+          in: query
+          description: Column to order by. Defaults to created_at.
+          schema:
+            type: string
+            enum: [created_at, updated_at, last_accessed_at, access_count, importance]
+            default: created_at
+        - name: order
+          in: query
+          description: Sort direction. Defaults to desc (newest / highest first).
+          schema: { type: string, enum: [asc, desc], default: desc }
       responses:
         "200":
           description: OK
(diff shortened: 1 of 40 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-07-12 09:32 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/96#issuecomment-12005)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #101), which is the source of truth. Further review and merging happen there.

PR #39, opened 2026-07-11
Let the reranker see more candidates than the recall limit

perfectra1n:
## The problem

`finalizeRecall` hands the reranker exactly the `k` memories it is about to return:

```go
pool := search.Dedup(ranked, k)
```

So the reranker can reorder the result set, but it can never change which memories are in it. If the vector and keyword legs put the genuinely best answer at rank 11, the cross-encoder never sees it and it never comes back.

That is backwards from what a reranker is for. Retrieval is cheap and approximate, and it is supposed to over-fetch so that something more accurate can make the final call. Recall already does the over-fetching: `poolK = max(k*poolFactor, poolFloor)`, so 50 candidates per leg are sitting right there in memory by the time we get to `finalizeRecall`. We just throw them away before the one component that could actually judge them gets a look.

A cross-encoder is a better judge than the fused score precisely because it reads the query and the document together instead of comparing two vectors that were produced independently and never saw each other. Restricting it to reordering the top `k` spends the whole cost of that model on the easiest part of the problem.

## What this changes

Adds `MEMINI_RERANK_POOL`. The reranker gets the top N composite-ranked candidates, and the result is truncated to `k` afterwards (which `finalizeRecall` was already doing at the end). The candidates are already retrieved, so the extra depth costs reranker time and nothing else. No second search, no extra store round trip.

It defaults to `0`, which reranks exactly the limit, so this is a no-op until someone opts in. I did not want to quietly multiply anyone's rerank latency by 5x on upgrade.

That said, I think the default is worth a conversation. Reranking only the limit is close to a no-op in retrieval terms, and I suspect most people who set `MEMINI_RERANK` assume they are getting the rescue behaviour. If you would rather it defaulted to the retrieval pool size when a reranker is configured, I am happy to change it. I just did not want to make that call unilaterally in a PR.

## Also: the `MEMINI_RERANK_MAX_BATCH_CHARS` docs are misleading, and it bit me

The comment says to size it "just below the model's effective context", and the default of 6000 follows from that reasoning. But a Cohere-style `/rerank` server scores each (query, document) pair in its own forward pass. The model's context bounds a single pair. It never sees the batch as one sequence, so the batch total has nothing to do with the context window.

What `MaxBatchChars` actually is, is an HTTP payload guard. And sized as if it were a context guard, it is far too small: `CrossEncoder.Rerank` splits the pool into batches and scores them in a plain sequential loop, one HTTP round trip each. With a 50 candidate pool and realistic memory sizes, 6000 shards into more than a dozen serial requests. In my deployment the reranker lives across a WireGuard tunnel, so that reliably blew `MEMINI_RERANK_TIMEOUT` and fell back to composite order. The user gets a reranker that appears to be configured and is silently doing nothing.

This is much more likely to hurt than a large request body is, so I have corrected the comment to say what the knob really does and to warn that a low value shards a deep pool into serial round trips. I left the default at 6000 rather than change behaviour in the same PR, but it is worth revisiting, especially now that a deep pool is possible.

## Where this came from

I run memini against a local Qwen3-Embedding, and recall was returning its true hit at around 0.9 followed by a tail of results at 0.25 to 0.46 that were simply noise (`MEMINI_RECALL_MIN_SCORE` does not filter them, since it gates a min-max normalised score). Adding a cross-encoder was the obvious fix, and that is when I noticed it could only ever reorder the ten results I already had.

## Tests

Three new tests in `internal/service/rerank_test.go`:

- the reranker is handed more than the limit when a pool is set, and can therefore surface a memory that was ranked outside it
- the pool is a ceiling on what the reranker sees, not on what recall returns
- with no pool configured, it still sees exactly the limit, so existing deployments are unchanged

The first one fails on `main`, which is the behaviour this PR is about. Full suite, `go vet` and `golangci-lint` are clean.


eleboucher:
> _Originally by **eleboucher** on 2026-07-12 09:32 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/97#issuecomment-12004)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #100), which is the source of truth. Further review and merging happen there.

PR #40, opened 2026-07-12
Generate the docs reference from the code, and restructure docs around it

perfectra1n:
> **Depends on #38 and #39.** This branch is stacked on both, so the diff below includes their commits until they land. Once they merge, this reduces to the five `docs:` commits. Both PRs add settings (`MEMINI_RERANK_POOL`, `MEMINI_ACTIVITY_*`), and the generator here already picks them up, so nothing conflicts either way.

memini has grown a lot of knobs, and the docs have not kept pace. There are around sixty-five settings now, nine MCP tools, twenty-two CLI commands and thirty-odd endpoints, all documented (where they were documented at all) by hand in a 635-line README. This makes the reference generated instead, restructures the docs so there is a path from a question to an answer, and adds worked setups.

## Generate the reference from the code

The information is already in the code and it is good. Every field on `config.Config` has an env tag, a default and a doc comment, and the comments are usually better than the README table was. `internal/api/mcp` knows exactly which tools it registered. `api/openapi.yaml` is current. The docs were a second, hand-maintained copy, and a second copy of anything drifts eventually.

Four generators, all behind `mise run docs`, all gated in CI the same way `api.gen.go`, the chart docs and `go.mod` already are:

- **`docs/reference/configuration.md`** from the env tags, doc comments and `deprecatedVars`.
- **`docs/reference/mcp-tools.md`** by starting the MCP server over an in-memory transport and asking it what it registered. The tool schemas are built at runtime by a generic helper, so there is no literal for a parser to read, and asking the server directly cannot disagree with what an agent actually sees. It builds the server twice, with and without an answerer, which is how it works out that `memory_answer` only exists when an LLM is configured.
- **`docs/reference/cli.md`** from the cobra tree, through a build-tagged subcommand, since the tree lives in `package main` and cannot be imported.
- **`docs/reference/rest-api.md`** from the OpenAPI spec.

The part that does the real work is that the generator **refuses to run** when a setting has no doc comment, belongs to no section, or is read from the environment without being declared. Rendering whatever it happens to find would not have been enough. Seventeen fields had no doc comment of their own, so their descriptions lived only in the README table. Those descriptions now live on the fields, where `go doc` and any editor pick them up too.

If you want to check the gate is real: add an env-tagged field to `config.Config` without a doc comment, or without adding it to the group table in `cmd/gendocs/main.go`, and run `mise run docs`. It should refuse, and say what to do.

## Restructure

The README is now 153 lines and routes rather than explains. `docs/README.md` is the front door that was missing.

`scopes.md`, `tiers.md`, `categories.md` and `api-keys.md` stay exactly where they are, deliberately. `config.go` prints `see docs/scopes.md#knobs` in the two boot-fatal deprecation messages and `config_test.go` asserts that string, so released binaries name that path when they refuse to start. Moving it would break the one message an operator reads when they are already stuck.

New, hand-written:

- `guides/tuning-recall.md`, organised as symptom, then cause, then the setting. It opens by saying to run `memini doctor` first, because the most common cause of bad recall is a namespace mismatch rather than ranking.
- `guides/solo-laptop.md`, `guides/homelab-team.md`, `guides/multi-agent-namespaces.md`.
- `operations/upgrading.md`, which leads with "my server will not start", since the two boot-fatal variables and the `memini migrate scopes` fix previously existed only as Go string literals.
- `operations/web-ui.md`, `deployment.md`, `import-export.md`.
- `reference/env-vars.md`. Four names (`MEMINI_API_KEY`, `MEMINI_NAMESPACE`, `MEMINI_HOME`, `MEMINI_AGENT`) are read by both the server and the agent-side plugin and mean different things on each side. That is an easy trap and it was not written down anywhere.

## What this surfaced

Writing a generator meant diffing the docs against the code, which turned up a handful of things worth fixing on their own:

- **`MEMINI_DEMOTE_AFTER` was documented as `0`, meaning off. It defaults to `168h`.** Durable memories that were never recalled, are unimportant and are uncorroborated get demoted back to episodic after a week. It shipped opt-in and the default was changed later. This one is worth a look independently of the rest of the PR, since it is the answer to "why did my durable facts go away" and the docs said it could not happen. `docs/tiers.md` and `charts/memini/values.yaml` repeated it.
- `MEMINI_CONSOLIDATE_MIN_SCORE` was documented as `0.6`. It is `0.3`.
- Twelve live settings were missing from the table, and they were mostly the recall-tuning ones (`MEMINI_RECALL_MIN_SCORE`, `MEMINI_RECALL_SEMANTIC_RESERVE`, `MEMINI_STABILITY_K`, `MEMINI_EPISODIC_MIN_CHARS`, `MEMINI_TURN_ECHO_WINDOW`), which is probably why there was no good answer to "my recall is bad".
- The Web UI section listed five views. There are ten. It also described the UI as backed by three read-only endpoints, but it does key CRUD and rotation, namespace moves and splits, reassignment and deletion.
- `MEMINI_AGENT` is applied by the plugin's namespace resolution, not by the server's header-less fallback, so it does nothing under a bare `memini mcp`. I had this wrong in my own first draft of the generator, and so did the old docs.
- `MEMINI_ACTIVITY_LOG` is on by default. I had written that backwards too, and it only got caught because the generated table and the prose disagreed with each other.
- The `--min-length` import gate defaults to `20`, not off, so it quietly drops short records.
- `integrations/README.md` listed eight MCP tools (there are nine, `memory_history` was missing) and linked to a README anchor this change removes.

Also drops `honcho_steal_report.md`, 386 lines of internal planning notes that have been shipping at the repo root.

## Reviewing

The five `docs:` commits are meant to be read in order and each stands on its own. The generated pages are long, so I would skim `docs/reference/configuration.md` and read the hand-written guides properly. Happy to split any of it out if you would rather take it in smaller pieces.


eleboucher:
> _Originally by **eleboucher** on 2026-07-12 09:32 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/98#issuecomment-12003)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #99), which is the source of truth. Further review and merging happen there.

PR #41, opened 2026-07-13
Plugin settings introspection, a per-project namespace override, and a fix for the concurrent-session namespace race

perfectra1n:
### You can't see what the plugin is actually doing

The client is configured entirely through environment variables, about 25 of them, read by JS, TS and Python that `memini doctor` (Go) can't see. Nothing prints the effective values, and nothing shows where a value came from.

That second part turns out to be the whole game. The case that started this: on my machine `MEMINI_NAMESPACE=default` had been set as a fish *universal* variable (`set -Ux`, persisted in `fish_variables`) months ago and then forgotten. Every repo on the box therefore resolved to one shared `default` namespace. A dozen projects' memories had piled into a single pool, recall was noisy, and the agent kept "remembering" things from unrelated work. Nothing about the symptom points at the cause.

`doctor` has had a `warnGlobalNamespacePin` check for exactly this. But `doctor` needs the `memini` binary, and a plugin-only install pointed at a remote server doesn't have one. The people most likely to hit this are the least equipped to find it.

### Concurrent sessions race on a single global file

This is the real bug, and it's worse than the first one.

The MCP `headersHelper` is the only thing that sets `X-Memini-Namespace` for `memory_remember` and `memory_recall`, and it's handed almost nothing. I instrumented it against a live session:

```
cwd                : <plugin install root>
PWD                : <plugin install root>   <- rewritten; NOT the project
CLAUDE_PROJECT_DIR : unset
CLAUDE_PLUGIN_ROOT : <plugin install root>
process.ppid       : the session's `claude` process, cwd = the project dir
```

So `process.cwd()` and `PWD` are both traps. Resolving from either derives the namespace from the plugin's own version-named install dir (`0.6.7`). The helper instead fell back to a single global cache file that the SessionStart hook writes.

That file is shared by every concurrent session, and it's last-writer-wins.

Open two sessions in two repos and both send whichever namespace started most recently, while the hooks (which *do* get a `cwd`) keep writing to the correct one. The agent saves memories it can never recall. That's the same "writes land where recall doesn't look" split `doctor` exists to diagnose, except produced by the plugin itself.

It reproduced immediately. With three sessions open, the global file read `claude-code-sync` while the current session's hooks resolved `default`.

### There's no way to set a project's namespace

Your options were `MEMINI_NAMESPACE` (global, and the cause of the first problem), hand-editing `~/.config/memini/config.json`, or deleting `project-map.json`. There's no `memini namespace set`.

### `/memini:status`

```
NAMESPACE
  effective              default        <- env
  git/cwd would give     memini         <- git-remote
  home (personal)        (unset)

MCP TOOL CALLS
  header the helper sends   default

WARNINGS
  [!] global-namespace-pin: MEMINI_NAMESPACE is set to "default", which pins EVERY
      project on this machine to one namespace. This project would otherwise resolve
      to "memini". If this variable is exported from a shell rc (or a fish universal
      variable), every repo you work in is sharing one memory pool.
      fix: Unset MEMINI_NAMESPACE, or set a per-project override instead.
```

A plain list of values would show `namespace: default` and look fine. What catches the problem is the line under it. So `describeSettings()` resolves the namespace three times against progressively stripped inputs and reports all three: what's in force, what it would be without the override, and what pure git/cwd derivation gives.

It also cross-checks the three things that have to agree (what the hooks write, what the MCP tools write, and the read set the server assembles) and warns when they diverge. Secrets are always redacted (`sk-…4f2a`), so the output is safe to paste into an issue.

It lives in the plugin rather than in the Go binary, specifically so it works for the plugin-only remote installs that `/memini:doctor` has never worked for. Server state comes from `GET /v1/namespaces/read-set` and `GET /healthz?verbose=1`.

One note on that: `/healthz` is deliberately not the reachability probe. A remote memini usually sits behind an ingress that only routes `/v1` and `/mcp`, so `/healthz` 404s while the server is perfectly healthy. I confirmed that against a live deployment. Reachability comes from the read-set call, which we need anyway.

### Fixing the race: use the process tree

The helper's parent *is* the session's `claude` process, and that process's cwd *is* the project. The hooks land on the same ppid, because Claude Code runs them via `sh -c '…run.sh script'` and `run.sh` execs node, which preserves the parent. So both sides key off the same value without having to coordinate:

```
1. CLAUDE_PROJECT_DIR         if it's ever provided
2. the parent process's cwd   /proc on Linux, lsof on macOS. Always fresh, and works
                              on the first connect, before any hook has run
3. sessions/pid-<ppid>.cwd    portable fallback (Windows), written by the hooks
4. the legacy global file     last resort, back-compat
```

Every step yields a *directory*, never a namespace. Re-resolving on each connect is what lets an override take effect on reconnect; a cached namespace would go stale the moment someone set one.

Both `exec`s in that chain are load-bearing. Drop either one and it becomes `claude` to `sh` to `node`, so node's parent is the shell (whose cwd is the plugin root) and the race quietly comes back. Nothing else in the suite would notice, so there's a test pinning it.

Two more things fell out of measuring this:

* The helper called bare `node`, while every hook carefully goes through `run.sh`. Under a node version manager (mise, nvm, fnm) it would simply fail. It uses `run.sh` now.
* `CLAUDE_PLUGIN_ROOT` *is* passed to the helper, so `.mcp.json` no longer depends on the `plugin-root` cache file existing. That file only exists once SessionStart has run, so the old path was broken on a cold first connect.

### `/memini:namespace`

```
/memini:namespace                # show the namespace and where it came from
/memini:namespace acme/api       # override it for this project
/memini:namespace --clear        # back to automatic resolution
```

The override beats `MEMINI_NAMESPACE`. That ordering is the point: if the environment won, this command would silently do nothing on exactly the machines that have the problem.

Two limitations, both of which the command states rather than hides:

* The hooks pick it up immediately, but the MCP tools need `/reload-plugins`, because Claude Code only runs the `headersHelper` when the server connects.
* Scope is the project (keyed by the git root), not the session. The `headersHelper` gets no session id, so it can't tell two sessions in one repo apart. Per-project is the honest granularity. Pretending otherwise would ship a setting that looks like it works and doesn't.

`memini doctor` reads the same file with the same precedence, so the two can't disagree about which namespace is in force. That seemed important for a diagnostic.

### Four more commands

`/memini:remember`, `/memini:recall`, `/memini:pin`, `/memini:forget`. These drive the existing MCP tools instead of duplicating REST. Two sharp edges are baked into their instructions:

* `memory_update` replaces the tag list rather than merging into it, so pinning without first reading the current tags silently destroys them.
* `memory_forget` is a permanent delete, so it confirms first, and suggests `memory_update` when a memory is merely wrong rather than unwanted.

### `MEMINI_SESSION_DIGEST=0`

Session digests are activity records ("edited `auth.go` (3), ran `go test ./...`"), not knowledge. They answer "what was I doing in this repo last week", which some people want and some emphatically do not: if you only want memini to hold durable facts, every session otherwise adds a memory that will never answer a question and quietly dilutes recall.

One switch covers all four write sites, since they're the same distilled buffer: the SessionEnd digest, the Stop checkpoint, the PreCompact rescue copy, and the PostToolUse buffering that feeds them (with digests off nothing would ever read that buffer, so the hot-path write is skipped too). Independent of `MEMINI_CAPTURE_TURNS` and `MEMINI_INLINE_EXTRACT`, which are easy to confuse it with.

## The model didn't know what it could do

Separately, I audited what the prompts actually tell the model against what the code actually supports. The MCP tool surface is in good shape and current with the cascade model. The gaps were everywhere else, and some of them were real bugs.

**A brand-new project got no memory directive at all.** `SessionStart` returned early when the briefing came back empty, *before* emitting `MEMORY_INSTRUCTION`. So the agent was told to save durable facts in every session except the ones where nothing had been saved yet, which is precisely when it matters. Same on an unreachable server: that should degrade to "no context", not to "and also stop saving".

**`MEMORY_INSTRUCTION` never mentioned `visibility`.** It's the only memini text injected into *every* session, and the default is `project`. So every user preference the agent captured was stranded in whatever repo it happened to be in instead of following the user. One line fixes it.

**The briefing dropped `scope_header`.** The MCP server's own instructions tell the model to name an ancestor namespace for `visibility` by reading it "off the briefing Scope line", and `memory_remember`'s error message enumerates the same chain. The hook was never showing that line, so `visibility: "<ancestor>"` was effectively unreachable through the hook briefing.

**pi, openclaw and hermes register their own tool schemas, and those were pre-cascade.** They don't proxy MCP; their recall schema was `{query, limit, tags, metadata}` and remember was `{content, id, tier, tags, category}`. No `scope`, no `visibility`, no briefing. PR #36 updated `plugin/skills/` but never touched the integrations, so a model on those harnesses could never deliberately read an ancestor namespace or save a memory that follows the user. Those are the two levers the whole cascade model is built around.

All three now get `scope` on recall, `visibility` on remember, and a `memory_briefing` tool, with descriptions lifted from `serverInstructions` so the model gets one story across harnesses rather than four dialects. Two details worth flagging, because without them the new levers would have been decorative:

* **Read provenance survives.** Results now carry `namespace` and `from`. The descriptions tell the model to read them, and stripping them made that a lie.
* **A rejected write reaches the model.** All three degraded a non-2xx to `null` plus a stderr warning. But the server's error on a bad `visibility` is exactly what enumerates the valid ancestor chain, and that error is the model's teacher. The remember path no longer swallows it. It still never throws into the host.

An unknown `scope` is dropped rather than forwarded, since `/v1/search` 400s on one and a model hallucinating `subtree` shouldn't turn a recall into an error.

**openclaw's memory skill told the model to type a raw namespace header.** "Set it to a project name (e.g. `acme-web`)" is the exact anti-pattern #36 removed. Rewritten around `scope`/`visibility`.

Rewriting it turned up something worse. The skill's curl recipes send the base `$MEMINI_NAMESPACE`, but the plugin defaults `namespace_per_agent` to true with template `{namespace}-{agent}`. So the plugin reads and writes `openclaw-alice` while a model following the skill writes to `openclaw`. **The two never meet**: a fact saved through the skill lands in a namespace the plugin's own recall will never look at, and simply disappears. Both calls return success, so nothing surfaces it.

The skill now leads with the tools and refuses to compete with them: if the `memory_*` tools are in the tool list, use those and ignore the file. The curl path stays, because the skill is also the documented way to run memini *without* the plugin, where a shell is all the model has. It just no longer pretends the two are interchangeable.

### `expose_tools` now defaults to on (behavior change)

Worth reading carefully, because it changes behavior for existing openclaw users.

`expose_tools` defaulted to `false`, so a plugin user got **no memory tools at all** unless they opted in. The reasoning was that the memory slot already recalls and captures automatically, so explicit tools were redundant.

That reasoning does not survive this PR. The automatic slot recalls with default scope and captures as `project`. It cannot express `scope: "everywhere"`, cannot write `visibility: "personal"`, and cannot ask for a briefing. Those are exactly the levers the cascade model is built on, and with `expose_tools: false` the model had no way to reach any of them. Combined with the skill bug above, an openclaw user's agent had *no correct path* to an ancestor namespace or a personal memory.

So the default flips to `true`. Set `expose_tools: false` to restore the pre-existing slot-only surface.

**`memory_history` existed and no prompt anywhere mentioned it.** Neither did `as_of` or `query_rewrite` in most places. Bi-temporal lineage is a headline feature nobody would ever have invoked.

**Skills pointed at `docs/categories.md`, which the plugin doesn't ship.** For a user in any other repo that path resolves to their own nonexistent file, so the model would try to read it, fail, and guess. Inlined.

## `stored: true` when nothing was stored

`memory_remember`'s contract said "stored=false means a low-signal write was dropped by the value gate", which implies the converse. Two paths break it, and both return a non-nil memory so no caller could tell:

* the exact-restatement fingerprint fast path, where a repeat of content already in the tier reinforces the existing memory instead of writing
* `MEMINI_WRITE_DEDUP_ACTION=coalesce`, where a near-duplicate that isn't richer than the stored one is folded into it and dropped

In both cases the agent is told `stored: true` and handed an id. It believes it saved a new fact and tells the user so. Worse, that id belongs to a memory it never wrote, so a follow-up `memory_update` or `memory_forget` clobbers someone else's memory rather than its own.

There's now a `reinforced` flag on the result (MCP and REST), set on both paths, and the tool description spells out what it means. While in there: `auto_superseded` was already being returned but explained in no description at all, so the model got a boolean it had no way to interpret even though it means a prior memory was just tombstoned.

## On drift between the plugins

Worth being explicit about, since it was a goal.

Namespace resolution was implemented five times before this (`plugin/scripts/_shared.mjs`, `packages/namespace-resolver`, opencode's `memini.js`, hermes' Python, `internal/config/namespace.go`). Bolting an override and a settings dump onto each would have made it seven.

So there's a new `packages/memini-client`: override, settings introspection with provenance, redaction, project-dir recovery. Who consumes it, and who doesn't:

* **Claude Code and Codex** import `plugin/scripts/_client.gen.mjs`, a committed esbuild bundle. The plugin ships as raw files with no install-time build step, which is the same constraint that made those hooks `.mjs` and not `.ts` in the first place. `mise run client-check` fails CI on drift, mirroring the existing `docs-check` gate. The bundle is byte-deterministic, which is what makes that gate meaningful.
* **pi and openclaw** inline it via esbuild `--alias`. They can't take it as a normal dependency, because both get `npm install`-ed by their hosts and an unpublished workspace package would make them uninstallable. Their own bundle tests already enforce that.
* **The Go CLI** reads the same `overrides.json`.
* **opencode** reimplements the reader, about 50 lines. It ships from npm as a single dependency-free file the host installs with Bun. Adding a bundler to save 50 lines of `JSON.parse` and `git rev-parse` would trade away the one property that makes it easy to install.
* **hermes and openwebui** reimplement it too. They're Python, so the TS core is unreachable.

Drift is gone where a shared artifact is possible (five surfaces). Where it genuinely isn't (Python, and a deliberately dep-free npm package), the file format is the contract instead. Which is why that format is boring on purpose: JSON, one `git rev-parse` for the key, and every error path degrading to "no override" rather than raising.

`memini-client` deliberately does not depend on `@memini/namespace-resolver`. They're siblings: that one is a resolution chain, this one is override plus reporting. Each harness already has its own resolver (the Claude hooks consult a self-healing project map the shared resolver knows nothing about), so `describeSettings` takes the harness's resolver as a callback instead of imposing one. Bundling a second, subtly different resolver into a plugin that already has one is precisely the drift this is meant to end.

### One thing I deliberately didn't change

openclaw's default namespace. It defaults to the literal `openclaw` and does no git derivation on purpose, because it's a gateway harness where `cwd` is often meaningless. Changing that default would silently relocate every existing user's memories.

The actual bug there is narrower, and it's fixed: openclaw ignored `MEMINI_NAMESPACE` *and* the override entirely (its namespace was just `c.namespace || "openclaw"`). `namespace_prefix`, `namespace_template` and per-agent nesting still apply on top, unchanged.

## A note on the Windows fallback

Worth calling out because it's the one place this design is weakest, and it came out of review rather than out of writing it.

The process-tree lookup needs `/proc` (Linux) or `lsof` (macOS). On Windows there's neither, so the pid-keyed session file is the *only* mechanism. A bare pid is not a durable identity: Windows recycles them quickly, and `fs.existsSync` on a recorded path is not a freshness check, because the old repo still exists on disk. It's just the wrong repo.

So: the record carries a `writtenAt` and is refused past a 6h TTL (and refused on a backwards clock, which would otherwise sail straight through an `age > TTL` check); SessionEnd deletes it, so a clean exit leaves no window at all; and Stop refreshes it, which fires once per assistant turn, so any session actually in use stays comfortably inside the TTL. That's what lets the TTL be short enough to bound pid reuse without ever expiring under a live session. Only a crash leaves a record behind, and the TTL bounds that.

## Verification

The cross-harness contract is tested rather than assumed. An override set with the Claude Code plugin reads back correctly from `memini doctor`, the shared TS core, pi, openclaw, opencode and hermes. All six report `source: override`.

* Go: `build`, `vet`, `gofmt`, `go test ./...`, `golangci-lint`, all clean
* `@memini/client`, Claude Code and Codex hooks, pi, openclaw, opencode, hermes, openwebui: all pass
* generated-docs and `client-check` drift gates: clean

The new tests target the things that are invisible and easy to break later:

* two parents in two repos getting two namespaces (the race itself)
* the `exec` chain preserving the parent, because dropping either `exec` silently restores the race and nothing else in the suite would notice
* an expired session record being refused, and a clean exit deleting it
* an override beating a globally pinned `MEMINI_NAMESPACE`
* an empty briefing still emitting the memory directive
* `MEMINI_SESSION_DIGEST=0` reaching all four write sites, not just the obvious one
* `reinforced` on both the fingerprint and coalesce paths, and *not* on a genuinely new write
* CR/LF rejection in a namespace, which matters because a namespace travels as an HTTP header, so client-side validation is deliberately stricter than the server's: a `\r\n` would split `X-Memini-Namespace` and let a caller inject arbitrary headers


perfectra1n:
GitGuardian, you smell

eleboucher:
> GitGuardian, you smell

don't be mean with my first company

eleboucher:
> _Originally by **eleboucher** on 2026-07-13 07:12 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/110#issuecomment-12207)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #111), which is the source of truth. Further review and merging happen there.

PR #43, opened 2026-07-13
Configuration redesign: server handshake, pins, per-key settings, admin access control, and a management UI

perfectra1n:
# Configuration redesign: server handshake, pins, per-key settings, and a management UI

## The problem

Configuration had grown past the point where I could hold it in my head, and I wrote most of it. The client side alone had about 30 `MEMINI_*` env vars plus six files on disk (`overrides.json`, `config.json`, `project-map.json`, a legacy namespace cache, a plugin-root pointer, and per-session cwd records). Namespace resolution was implemented four times with subtle differences: the Claude plugin's `_shared.mjs`, the `@memini/namespace-resolver` package, the Go CLI, plus hand-rolled override readers in opencode, hermes, and openwebui because they cannot import TypeScript. Four env var names (`MEMINI_API_KEY`, `MEMINI_NAMESPACE`, `MEMINI_HOME`, `MEMINI_AGENT`) meant different things on the server and the client, so exporting one in a shell configured both sides differently. `MEMINI_MCP_URL` was documented, listed in the settings introspection, and consumed by nothing.

The deeper issue: the server computed a `Principal` (key name, bound home namespace, default namespace) on every request but never told the client. So the client re-derived everything locally, `/memini:status` could only guess at split-brain between the hooks and the MCP helper, and a `visibility:"personal"` write errored on any machine that never exported `MEMINI_HOME`, even when the key was bound to a home server-side and the server knew better.

## The fix, conceptually

Invert control. At session start the client sends project facts (git remote, toplevel, cwd basename, agent, its `MEMINI_NAMESPACE` if set) plus its API key to a new endpoint, and the server answers with everything: the resolved namespace with provenance, the caller's identity, and the behavioral settings. Settings become data on the server, editable in the UI and by API, following the key across machines, instead of env vars scattered through shell rc files.

Config now has four layers, and each one has exactly one home:

| Layer | What | Where it lives |
|---|---|---|
| Bootstrap | how to reach the server | client env (`MEMINI_BASE_URL`, `MEMINI_API_KEY`, `MEMINI_REQUIRE_HTTPS`, `MEMINI_DEBUG`, `MEMINI_AGENT`) |
| Identity | who you are | your API key, learned via the handshake |
| Context | which project | client-gathered facts, resolved by the server (pin beats env beats derivation) |
| Behavior | how the client acts | server-stored settings; a client env var still wins when explicitly set, as a debug override |

```mermaid
sequenceDiagram
    participant H as Hook / MCP helper
    participant S as memini server
    H->>H: gather facts (git remote, toplevel, cwd, agent, env ns)
    H->>S: POST /v1/handshake {facts} + API key
    S->>S: pin? env? declared? derive; merge settings (builtin < global < key)
    S-->>H: namespace + source, identity (key, home, default ns), settings + sources
    H->>H: cache per session (pid-keyed, 10 min TTL, cwd + facts validated)
```

### 1. `POST /v1/handshake` and the `internal/nsresolve` package

Why it was a problem: four resolver implementations drifted, and the client and server could disagree about which namespace a project maps to with no way to detect it.

Why this way: resolution is a pure function of facts, so it belongs in one place, on the server, where pins and per-key policy live. The client keeps only a minimal fallback (same derivation order) for when the server is unreachable, and since both sides implement the same chain from the same test vectors, degraded results converge with the server's derived answers.

What we did: a transport-free `internal/nsresolve` package resolves pin > client env > declared > derive (remote name or owner_repo slug, then prefix and agent suffix) > key default > server default. The handshake is deterministic and side-effect free. A shared fixture (`derivation-vectors.json`) is consumed in place by the Go, TypeScript, and Python test suites, so the derivation logic cannot fork again without a test failing somewhere.

### 2. Pins replace `overrides.json` (`/v1/pins`)

Why it was a problem: the per-project override lived in a local file, so it silently did not follow you to your other machines, and every non-TS integration had to reimplement the reader.

Why this way: a pin is deliberate, attributable intent about a project, which is exactly the kind of thing the server should store. Keying stays the proven dual scheme from the old client cache: `remote:<canonical-url>` for cross-machine identity and `path:<toplevel>` for remote-less repos. Only explicit pins are stored; derived answers are computed per request, so the table cannot fill up with stale garbage. The old rule that the override beats `MEMINI_NAMESPACE` carries over unchanged (a forgotten global export must not silently defeat `/memini:namespace` on exactly the machines that have the problem).

What we did: a `project_map` table behind a `ProjectMapStore` capability (both drivers), CRUD at `GET/PUT/DELETE /v1/pins`, writes allowed for any caller and activity-logged with the author as the control. `POST /v1/namespaces/move` renames pins along with everything else. `/memini:namespace` now manages pins, and session start migrates an existing `overrides.json` entry to a pin automatically, once, with a one-line notice. `/memini:namespace --migrate` does the bulk version and renames the file when everything lands.

### 3. Per-key settings and `MEMINI_CLIENT_DEFAULTS`

Why it was a problem: capture flags and injection budgets (15+ vars) lived only in per-machine shell env. Invisible to the server, invisible to `status` on any other machine, and not manageable for a key you handed to an agent.

Why this way: behavior is a property of the credential, not the shell. The `api_keys` table already carried per-key `home_ns` and `default_ns`, so settings extend that pattern. Layering is built-in defaults, then global defaults, then per-key, and a client env var still wins when set because a debug override you can flip in one shell is worth keeping.

What we did: a validated `ClientSettings` blob (23 fields) on `api_keys` plus a global defaults blob in the existing meta KV, merged with per-field source labels that flow through the handshake as `settings_sources`. File-managed keys can declare settings in the keys YAML, validated fatally at boot. For GitOps deployments, `MEMINI_CLIENT_DEFAULTS` (JSON in the server env, one line in Helm values) becomes the global layer and locks the API editor read-only with a 409. New self-service surface: `GET /v1/self` and full-replace `PUT /v1/self/settings` (a named key edits its own settings without admin). Admins edit any key through the extended keys API, and the UI exposes all of it.

### 4. Client core and the caching discipline

Why it was a problem: the old MCP headers helper fell back to a single global namespace cache file, which caused the concurrent-session race fixed in the previous release. Any new caching had to not reintroduce that class of bug.

Why this way: cache per session or not at all. The handshake result is cached in `sessions/pid-<ppid>.handshake.json`, validated on read by TTL (10 minutes), exact cwd match, and a facts fingerprint, so pid reuse and mid-session repo switches invalidate naturally. Two concurrent sessions physically cannot read each other's entries. The legacy global cache is not just unwritten, its read path is deleted; a stale file on disk is ignored.

What we did: new modules in `@memini/client` (bootstrap env reading, fact gathering, the local fallback resolver, the handshake wire client and cache), bundled dependency-free into the plugin. Hooks follow a strict network policy: session start does a live handshake, stop and session-end use cache-then-live, and the per-tool-call hooks are cache-only and never touch the network. When the server is down everything degrades to local derivation with the degradation shown in provenance, and the session keeps working.

### 5. Plugin hooks, `/memini:status`, and split-brain

Why it was a problem: hooks and the MCP helper resolved namespaces independently, so they could disagree, and `status` existed largely to catch that after the fact.

What we did: both now ask the same deterministic endpoint, so a namespace split can only come from different input directories; `status` checks exactly that (facts fingerprint comparison) and otherwise reports what the server said, with per-setting provenance (`<- env (overriding server)`, `<- server (key)`, `(default)`), the caller's identity, and a clear degraded-mode warning when the handshake failed.

### 6. Integrations

Why it was a problem: opencode, hermes, and openwebui each reimplemented the override reader and config templating (about 150 lines each in the Python ones) because the contract was a file format they could not import.

What we did: the contract is now a wire call. All three lost their override/template machinery and gained a small memoized handshake client (fail-soft, 2.5s timeout, 10 minute TTL). openwebui's admin valve rides as `declared_namespace` and its per-user suffix still applies client-side. pi and openclaw, which import the TS package, moved onto the same core modules, and `packages/namespace-resolver` is deleted outright. The Go CLI dropped `override.go`, mirrors resolution by importing `internal/nsresolve` (not by reimplementing it), and `memini doctor` now probes the handshake, explains local-versus-server differences by provenance instead of alarming, and flags leftover dead files.

### 7. Management UI

Why: settings you cannot see are settings you cannot debug, and this redesign moves a lot of state server-side.

What we did: a new Config view with four tabs: every setting with description, value, and provenance (identity panel included, admin key-picker for inspecting any key); a global defaults editor that renders read-only with a banner when `MEMINI_CLIENT_DEFAULTS` manages the layer; a pins table; and a resolution preview that calls the handshake with arbitrary facts to answer "why did I get this namespace". The Keys view gained a per-key settings editor (file-managed keys render read-only). Field descriptions are generated from the OpenAPI schema at build time so UI help text cannot drift from the contract.

### 8. Duplicate recall injections stop repeating themselves

Why it was a problem: the PreToolUse hook injects per-file memories on every Read, Edit, or Write. Working a file repeatedly injected the identical memory block over and over, burning context on content the conversation already had.

What we did: the hook fingerprints what it injects (the file plus the full untruncated content of the served memories, deliberately ignoring which tool triggered it) and skips the injection when it would byte-repeat the last one for that file. The recall itself still runs, because results can legitimately change between edits, including in-place `memory_update` edits whose diff lands past the display truncation, which the fingerprint therefore ignores. State is per-session, bounded, and cleared at session start, before compaction (a compacted context lost the earlier injection, so re-injecting is correct), and at session end. It is governed by a new `inject_dedupe` setting (default on) that runs through the entire pipeline this PR builds, which made it a nice proof of the design: one schema field, and the store, handshake, hook gate, docs, and the UI's settings form all picked it up, the last one with zero hand-written UI thanks to the generated catalog.

### 9. Naming cleanup

URLs and wire enums no longer contain dashes: pins live at `/v1/pins`, the read-set endpoint moved to `/v1/namespaces/readset`, and enum tokens are `key_default`, `server_default`, `owner_repo`. HTTP header names keep their conventional dashes.

### 10. Admin becomes a key attribute, and the UI gets a real login

Why it was a problem: "admin" was not an attribute anywhere, it was the absence of a principal. The env `MEMINI_API_KEY` and dev mode were indistinguishable, and a named key could never manage keys by construction. Worse, the UI's idea of login was the server embedding the admin env key into the public HTML shell as a meta tag, which handed the admin credential to anyone who could GET the page. Views detected adminness by string-matching a 403 message.

Why this way: admin should be a capability you grant to a credential, not a side effect of which credential type you hold. And a browser should prove it holds a credential, not be handed one.

What we did: API keys (table and file) carry an `admin` boolean, surfaced through the keys API, the keys YAML, and a `key add --admin` CLI flag. The gate became `requireAdmin`: the env key and dev mode still pass (bootstrap and break-glass are preserved), and a named key passes when admin is set. A named key cannot demote, disable, or delete itself (409), so you cannot saw off the branch you sit on; rotating yourself stays allowed because the new secret is returned to the very caller who proved possession of the old one. Grants and revokes land in the activity log. The identity endpoints now report an `admin` claim, and the UI runs on it: the meta-tag injection is gone, an auth gate verifies against `GET /v1/self`, a login screen persists the key after it verifies (never before), Settings gained a session panel with logout, and the admin views render locked states that name the signed-in key. Two session-survival details mattered: rotating your own key and creating the first key in dev mode both auto-adopt the returned secret before any refetch, so the one-time secret modal survives the auth flip. The dev-mode bootstrap defaults the first key to admin and warns if you uncheck it, because a non-admin first key would lock the browser out the moment auth turns on. The new 403 text ("admin credential required") is deliberately different from the old one so any out-of-tree matcher fails loudly instead of half-working.

### 11. The activity feed says who did it, and why

Why it was a problem: activity events recorded what happened but not who did it. The caller's identity was fully resolved at the middleware and then dropped before every read-path service call. The only "why" on a recall was its free-text query, write events recorded none of their decisions, and the pin and settings event kinds were invisible in the UI entirely.

What we did: every event row now carries an actor (the key name) and an actor kind (named key, env key, or open access), stamped once at a single funnel that every event writer flows through, so no current or future event kind can miss attribution. Recalls carry a structured `source` (the plugin's file-triggered recalls send `pretool`, the MCP tool stamps `mcp`, the UI sends `ui`, answer grounding sends `answer`, and anything else defaults to `api`); unknown values pass through untouched so no client can ever 400 on it. Write events record what the write decided (final tier, auto-supersede, merge hints). The Activity page shows actor and source chips, renders the config event kinds it previously hid, tolerates legacy rows that predate the columns, and gained an actor filter that threads down to the store query.

## Breaking changes (pre-1.0, deliberate)

- Removed client env vars: `MEMINI_URL`, `MEMINI_TOKEN` (aliases), `MEMINI_MCP_URL` (was inert), `MEMINI_NAMESPACE_SCOPE` (now the per-key `namespace_scope` setting). Setting them prints one ignored-vars notice at session start.
- `~/.config/memini/overrides.json` and `config.json` tenant templating are retired. Overrides migrate to pins automatically (or in bulk via `--migrate`); tenant roots are replaced by `namespace_prefix` on keys, with instructions printed by the migrator.
- `/v1/namespaces/read-set` renamed to `/v1/namespaces/readset`.
- One capability regression, on purpose: no offline pinning. Pins live on the server; `MEMINI_NAMESPACE` remains the offline escape hatch.
- Deleted package: `@memini/namespace-resolver`.
- The UI no longer auto-logs-in. The server used to embed `MEMINI_API_KEY` into the HTML shell; that injection is removed, and every browser logs in once through the new login screen (the key persists in localStorage). Upgrading.md walks each deployment style through it.
- The admin gate's 403 text changed from "admin key required" to "admin credential required", on purpose, so any out-of-tree matcher fails loudly rather than half-working.
- Pin and unpin activity events no longer duplicate the actor into `detail.by`; the event row itself now carries `actor` and `actor_kind`.

## Testing

Three tiers, all executed against this branch.

Unit: the shared derivation vector fixture runs in the Go, TypeScript, and Python suites, so cross-language resolution parity is a test rather than a convention. Handshake determinism and its no-writes guarantee are asserted (globally, not per-namespace). The authz matrix for all eight new endpoints is table-tested including the 403, 409, and 501 paths, across every credential class. Hook tests include a two-session concurrency regression, a stale-legacy-cache-ignored regression, the same-session migration invariant, and the dedupe matrix. Suites at HEAD: full Go, plugin hooks (95), client core (74), opencode (46), hermes (57), openwebui (36), pi (40), openclaw (139), UI typecheck and build, golangci-lint, docs and helm drift checks.

Integration: the store conformance suite (ProjectMapStore, ClientSettingsStore, per-key settings on api_keys, unknown-field tolerance) ran against a real VectorChord Postgres container, 77 tests passing, in addition to the default sqlite driver coverage.

The admin and activity work added its own coverage on all three tiers: the gate matrix runs every gated endpoint against every credential class (env key, dev mode, named table admin, named file admin, named non-admin), the self-guard 409s and rotate-self preservation are pinned, identity claims are asserted per class on both /v1/self and the handshake, attribution threading is tested for every event kind with the actor filter verified in both drivers, and a second live end-to-end pass covered the bootstrap flow (dev mode to first admin key with the session surviving the auth flip), all login flows including rotate-self, and the activity page showing who and why.

End to end: a live pass against a real server exercised the composed system: concurrent sessions with isolated caches, the full pin lifecycle across sessions, the settings cascade from admin defaults through per-key to env override, the `MEMINI_CLIENT_DEFAULTS` lock, migration idempotency including the same-session namespace guarantee, degraded mode with the server down, and the doctor's handshake probe. All seven scenarios passed. The pass also paid for itself: it caught that the plugin's briefing fetch still used a retired path-parameter route, invisible to every mocked test because the server's SPA catch-all answers unknown paths with HTTP 200. The route is fixed and its regression test now serves HTML on the old path, so that class of mock blindness cannot recur.

## Deliberate trade-offs and known follow-ups

Choices made on purpose, so reviewers know they are not oversights:

- Any authenticated caller can write a pin. The control is the activity log (pin and unpin events carry the author), not an authorization gate. Worth revisiting after 1.0 if teams want stricter ownership.
- A session that starts while the server is down writes to the locally derived namespace even when a pin exists for the project. The degradation is loud in provenance and reconciles once the server returns.
- `settings_sources` reports `global` whether the global layer comes from the API or from `MEMINI_CLIENT_DEFAULTS`; the defaults endpoint's `managed_by` field carries the distinction and the UI renders it. `/memini:status` does not currently make that distinction.
- On a gateway (openclaw), `MEMINI_NAMESPACE` in the daemon's environment beats even a pin, unlike everywhere else. A gateway env var is deliberate per-install configuration, not a forgotten global, so the usual pin-beats-env rationale inverts.

Smaller notes for future passes, none behavior-affecting: the store's multi-entry pin upsert test proves batch success but cannot force a mid-batch failure to prove rollback; hermes keys its handshake memo module-wide (documented single-provider assumption); an openwebui admin's valve edit can take up to ten minutes to apply while a cached handshake ages out; opencode accepts a server `recall_limit` of zero while hermes treats it as unset; a couple of UI polish items (tab arrow-key navigation, re-syncing a key's settings draft after save).


eleboucher:
> _Originally by **eleboucher** on 2026-07-14 07:39 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/113#issuecomment-12378)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #114), which is the source of truth. Further review and merging happen there.

PR #44, opened 2026-07-14
docs(chart): spell out the api-keys mount wiring and warn upgraders about the enabled flag

perfectra1n:
The persistence.api-keys block that landed in v0.7.1 defaults to enabled: false, and that default interacts badly with anyone who already carried their own persistence.api-keys block in release values from before v0.7.1. Helm merges the chart default underneath their block, enabled: false wins because their values never set the flag, and the keys mount and volume silently vanish from the rendered pod. The server then boots without the file and every named key stops authenticating. I hit this on my own deployment during the v0.6.9 to v0.7.1 bump; the rendered-manifest diff caught the StatefulSet losing the mount, nothing else did.

Comments only, no behavior change: the block's comment now walks through the three required pieces of the wiring (the Secret, enabled: true, and MEMINI_API_KEYS_FILE) and warns upgraders explicitly with the one-line fix; the env-side comment points at the enabled flag too, since setting either half alone does nothing. Chart README and schema regenerated with helm-docs and helm-schema; helm-lint and the drift checks pass.

I considered making the example fully commented instead (no merge trap at all), but the live block is what docs/guides/access-control.md builds on, so warning loudly seemed better than changing shape again one release later.


eleboucher:
> _Originally by **eleboucher** on 2026-07-14 17:35 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/115#issuecomment-12470)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #116), which is the source of truth. Further review and merging happen there.

PR #45, opened 2026-07-14
Make memory saving proactive: prompt sweep and event-aware auto-save nudge

perfectra1n:
## Why

Agents using memini recall well but rarely call `memory_remember` on their own. The save-side guidance was spread across several surfaces with vague triggers ("after learning something durable"), a deliberately conservative tone ("prefer quality over quantity"), and a Stop nudge that fired purely on user-message count, regardless of whether anything memorable happened, while inviting a no-op ("if nothing new is worth keeping, just stop again"). The recall side, by contrast, has three delivery layers (session briefing, per-file injection, tool description). In practice that asymmetry means the store grows mostly through automatic capture rather than deliberate saves.

Production memory prompts that do get frequent saves (ChatGPT's bio tool, Perplexity's memory system) share a few patterns: concrete trigger lists placed near the tool-choice decision, explicit "do not wait to be asked" language, save-before-acknowledge rules for phrases like "remember this", and a tight exclusion list that reduces hesitation on borderline cases. This PR applies those patterns to every memini prompt surface and makes the auto-save nudge mechanics event-aware instead of purely count-based.

## What changed

**Server prompt surfaces (`internal/api/mcp/mcp.go`)**

- The save bullet in the server instructions moves up to second position and now names concrete trigger categories: a decision and its rationale, a bug's root cause, a project convention, a stated preference or correction (a correction is a preference), an environment quirk, a non-obvious command. Explicit user requests save unconditionally, even if trivial or already stored (near-duplicates are reinforced or merged server-side), with secrets and credentials as the one stated exception. The exclusion list now includes secrets and credentials, which no surface mentioned before.
- The `memory_remember` description leads with agent-initiated triggers and trigger phrases, and tells the model to call the tool first and acknowledge after. Tier and result-flag documentation is unchanged.
- New correction-hygiene guidance in the instructions and the `memory_recall` description: when a stored memory proves wrong or outdated, fix it with `memory_update` or remove it with `memory_forget` rather than leaving known-incorrect data in place.
- A doc comment marks the server instructions as the canonical save-policy text and lists the mirror surfaces, so future edits keep the invariants aligned. `docs/reference/mcp-tools.md` regenerated.

**Plugin**

- `MEMORY_INSTRUCTION` (the SessionStart directive) is rewritten around the same trigger list, with a declarative-fact example pair ("User prefers concise replies" rather than "Always reply concisely"), the secrets exclusion, and the correction rule. The old "save nothing when nothing is worth keeping" paragraph is gone, and the tier guidance now matches the tool description (omit tier to auto-classify; it previously prescribed tiers, which had drifted).
- The Stop auto-save nudge is now event-aware. A pure decision function evaluates each interval boundary: it suppresses the nudge entirely when the transcript shows `memory_remember` or `memory_update` calls since the last baseline (saves made by subagents count), defers when fewer than `auto_save_min_events` tool events were buffered (trivial window), and falls back to a discussion-variant nudge at twice the interval for long conversations with no tool activity. When it does fire after real activity, the nudge names the actual files edited and commands run instead of a generic exhortation. The honest no-op path is kept: if nothing durable happened, the agent just stops again.
- After context compaction, the SessionStart re-fire appends a short recovery note asking the agent to save any durable facts that may have been compacted away before they are lost. Rides the existing `inline_extract` switch.
- The remember skill and command and the recall skill get the same treatment: expanded trigger lists, the secrets rule (stated to outrank the unconditional-save rule so the two never conflict), correction hygiene, and worked good/bad examples.
- `marketplace.json` said "5 hooks"; there are 6.

**Settings**

- New client setting `auto_save_min_events` (integer, minimum 0, default 3, env `MEMINI_AUTO_SAVE_MIN_EVENTS`). Setting it to 0 disables the activity gate and restores the old interval-only cadence. Threaded through openapi, the store types, generated Go and TypeScript artifacts, the client knob table, and the committed plugin bundle, mirroring how `inject_dedupe` is wired.

**Integrations**

- The pi and openclaw remember-tool descriptions are ported to the same text, adapted to their tool surface (they have no `memory_update`; corrections re-save with the existing id, and `memory_forget` handles removal).

**Docs**

- The plugin README documents the new heuristic and knob; `env-vars.md` gains the new variable.

## Bias calibration

The wording shift is deliberately moderate. There is no "when in doubt, save" anywhere: explicit user requests save unconditionally (secrets and credentials are the one exception), everything else needs a recognizable trigger, and trivial sessions defer the nudge. Bolder saving does produce more near-duplicates, but the server already absorbs those (value gate, reinforcement, merge hints, auto-supersede), and each surface now says so, which removes the hesitation without loosening the bar. If the reinforced or merge-hint rates in the activity feed spike after this lands, the triggers are too broad and can be tightened.

## Testing

- All suites green on the branch head: Go tests, plugin hook tests (110 to 113; new coverage for every row of the nudge decision table, save-observation suppression, defer plus discussion fallback, and compact recovery on both briefing paths), client tests, the bundle drift check, and the docs regen check.
- The heuristic lives in pure functions with direct unit tests; hook-level tests exercise the real scripts end to end against a mock server, including a test that embeds a `memory_remember` tool_use block in the transcript and asserts the nudge is suppressed.

## Note

`integrations/pi/plugin` fails `tsc` typecheck on main before this branch (postJson arity errors around `src/index.ts:1070`). This PR only touches a description string in that file; the esbuild build and its tests pass. Flagging it in case you want it fixed separately.


eleboucher:
> _Originally by **eleboucher** on 2026-07-14 19:52 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/117#issuecomment-12503)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #118), which is the source of truth. Further review and merging happen there.

PR #46, opened 2026-07-14
Tighten the vocabulary: one meaning each for namespace, project, pin, and scope

perfectra1n:
## Why

This came out of the "what do we actually call these things" conversation: project is used for the lower level, namespace acts as the tenant, and none of it is written down. Auditing the codebase confirmed the confusion is real but localized. "namespace" is the one true partition concept and is consistent everywhere it matters. "project" meant four different things: the scope/visibility enum value, the git repo identity behind pins, the admin UI's synonym for namespace, and a split grouping key. "tenant" survived only as prose left over from the removed `MEMINI_TENANT_SHARED` model. "scope" was overloaded on top of that (per-call read width, the subsystem name, the `namespace_scope` slug setting, and a UI view that actually shows the read set).

Rather than invent new words, this tightens the ones we have and writes the contract down.

```mermaid
flowchart LR
    P["project<br/>(client-side: the git repo a session runs in)"] -->|"pin > env > declared > derive"| N["namespace<br/>(server-side: acme/phoenix/api)"]
    N -->|"scope: project / full / everywhere"| RS["read set<br/>(primary + ancestors + home + links)"]
```

## What changed, by commit

**docs: glossary (new page).** One meaning per word: namespace (a node in the slash tree, the only partition key, server-side), project (client-side only: the repo/directory a session runs in, which resolves to a namespace), pin (the stored project to namespace binding), scope (per-call read width only), visibility (per-call write placement), plus read set/origin/source/home. It also disambiguates a pin from a pinned memory, and records the two deliberate blessings listed under trade-offs.

**docs: tenant retired as a live term.** README's "per-tenant isolation", the MCP tools' "tenant namespace" descriptions, the `X-Memini-Namespace` parameter description, code comments, and test fixtures now say namespace (or team/root/source where that was what was meant). Historical references stay where they name real things: the removed `MEMINI_TENANT_SHARED` knob, the retired `tenantRoots` client config detection, and the mnemory import format's `tenant` field.

**refactor(ui): the Projects view is now Namespaces.** The nav, view, filters, memory-card chip, and Health/Graph strings stop calling namespaces projects, and the Scopes nav item is now "Read set" since that is what the view renders. `/projects` redirects to `/namespaces` so old bookmarks keep working, and saved collapse state is folded forward from the old localStorage key. The pins and handshake surfaces keep saying project; there it is the right word.

**refactor(store): project_map is now pins.** The table, its index, and the Go surface (`store.Pin`, `store.PinStore`, `PutPins`/`GetPins`/`DeletePins`/`ListPins`/`RenamePinNamespaces`). The migration renames the table in place before the CREATE could plant an empty pins next to it, in both drivers. If a rollback-window binary re-created project_map and wrote pins into it, the next boot folds those rows into pins (the stray wins on a key conflict, the same last-write-wins rule as PutPins, while created_at/created_by keep the pins row's values) instead of stranding them, and the postgres rename tolerates a concurrent migrate winning the race between the guard and the ALTER. New migration tests in both drivers cover the rename and the fold.

**refactor(api): the ProjectMap* wire schemas are now Pin*.** Component names only: routes, operationIds, and every JSON property are untouched, so the generated server interface and the wire are identical. Verified by comparing `/v1/pins` responses from a main-built binary and this branch against the same database.

**test(config): rider.** `TestLoadDefaults` failed under lefthook's pre-push run from a linked worktree because git exports an absolute `GIT_DIR` to hooks, which resolved the repo from inside the test's temp dir. The test now clears `GIT_DIR`/`GIT_WORK_TREE` the same way `clearMeminiEnv` handles memini vars.

- `scope:"project"` and `visibility:"project"` are blessed, not renamed: on the agent-facing surface they read caller-relative ("this project's memories" is the namespace the caller's project resolved to), and renaming wire enum values would break every deployed client for zero semantic gain. The rationale is written into the glossary where the next contributor will find it.
- The per-key `namespace_scope` setting keeps its name: renaming it would need a permanent alias (settings PUT uses strict decode, and operators hand-author it in keys files), which re-creates exactly the duplication this PR removes. The glossary blesses it as "derivation slug style".
- `exact`/`subtree` stay as deprecated REST scope aliases, unchanged.
- The pins migration is one-way: a downgraded binary re-creates an empty project_map, and the next upgrade folds whatever landed there back into pins.
- The UI collapse-state localStorage key changed (`memini.collapsedTenants` to `memini.collapsedNamespaces`); saved state migrates once and the old key is removed.
- `api.gen.go`, `api-schema.gen.ts`, and `docs/reference` carry a comment-only fan-out from the spec description changes; those hunks are regen output, not hand edits.

## Verification

- `mise run test` / `lint` / `test-hooks` / `test-client` / `client-check` / `docs-check` / `ui` all green, and `go test -tags=integration ./...` green against a fresh vchord-postgres container (which runs the new postgres migration tests, including the scratch-database rename test).
- Real-binary upgrade check: a server built from main wrote a pin into project_map; this branch's binary, pointed at the same SQLite file, served the identical row from `/v1/pins` (same timestamps, so renamed rather than rewritten) and left only `pins` + `idx_pins_ns` on disk.
- UI driven in a real browser: the nav labels, the `/projects` redirect, the collapse-state migration (a legacy key holding the old `(no tenant)` sentinel folds forward and the box renders collapsed), and the top-bar namespace switcher still renders at its compact 12px mono. The `ns-box-*` class scoping exists because a first cut collided with the switcher's pre-existing `ns-name` class; review caught it before it shipped.


eleboucher:
> _Originally by **eleboucher** on 2026-07-14 20:40 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/119#issuecomment-12534)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #120), which is the source of truth. Further review and merging happen there.

PR #47, opened 2026-07-14
Make the client request timeout configurable (a slow reranker was capped at 5s)

perfectra1n:
## The problem

Reranking is effectively capped at **5s**, no matter how the server is tuned.

`MEMINI_RERANK_TIMEOUT` already bounds a rerank server-side (default `10s`), and `Service.finalizeRecall` degrades to composite order when it fires — so recall never fails on the reranker's account. But **every client hung up first**: the Claude/Codex hooks, openclaw, openwebui and hermes all hardcoded a 5s ceiling with no way to raise it.

So a cross-encoder that takes longer than 5s — a cold model, a deep `MEMINI_RERANK_POOL`, or a batch sharded into *serial* requests by `MEMINI_RERANK_MAX_BATCH_CHARS` — makes the client abort before the server can answer. Recall returns **nothing at all** rather than an unranked result, and the reranker presents as *"memory is broken"*, not *"memory is slow"*.

The stock defaults ship that way: **10s of server-side rerank budget behind a 5s client**.

Layered timeouts only degrade gracefully when the outermost one is the longest.

| Client | Before | After |
| --- | --- | --- |
| Claude/Codex hooks | `5000`, **no knob** | `MEMINI_TIMEOUT_MS`, default `30000` |
| hermes | `TIMEOUT = 5`, **no knob** | same knob |
| openclaw | `5000` + a raw `5000` bypassing its own `timeout_ms` | same knob |
| openwebui | valve `5000` | valve `30000` |
| pi / opencode | `MEMINI_TIMEOUT_MS`, `30000` | unchanged — already correct |

## The change

One knob, resolved at two layers, sharing one default and one floor:

- **Transport** (`bootstrap.ts`): `MEMINI_TIMEOUT_MS`. A request timeout *is* transport, and it must bound calls made **before** the handshake that fetches settings.
- **Server-pushed** (`BEHAVIOR_KNOBS`): `request_timeout_ms` on `ClientSettings`. The admin who deploys the slow reranker raises the ceiling **once, fleet-wide** (`PUT /v1/settings/defaults` or the admin UI), instead of asking every user to export a variable.

Precedence is the repo's existing `env-override > server > default`.

**Default `30000`** — the value `MEMINI_TIMEOUT_MS` *already* means in pi and opencode. Adopted rather than minting a second number, so the knob means the same thing whichever client resolves it (the rule `BEHAVIOR_KNOBS` itself states). It's a **ceiling, not a target**: the server still degrades at its own deadline, so a healthy client never waits near it.

`0` is deliberately **not** overloaded as "no timeout" (a client that never gives up hangs forever on a wedged server). The floor is `100ms`, and a below-floor value is *raised to it* rather than silently swapped for the default, so a deliberately tight timeout stays tight.

## Drive-by fix (separate commit)

While adding the field to the wire mapping I found **`auto_save_min_events` was never delivered to clients at all.** It's in the schema, the generated struct, `BEHAVIOR_KNOBS`, and the docs — but missing from *both* directions of the hand-written mapping in `config_shared.go`. `PUT /v1/settings/defaults` silently dropped it, and the handshake never returned it, so the server-side half of the event-aware auto-save nudge (#118) has never actually worked. Only the env var did.

These mappers are hand-written against a **generated** struct, so an omitted field fails **silently** — it compiles, round-trips as `nil`, and the setting just never applies. Nothing caught it because no test asserted a setting *crossing the wire*: the merge was tested (in `internal/store`) and the handlers were tested, but never the seam between them. `TestSettingsRoundTripAcrossTheWire` now pins that seam and fails without the fix.

## Tests

The invariant that caused the bug — **the client default must exceed the server's rerank budget** — is now pinned in both Go and TS, so it can't be silently re-inverted. Plus: the schema floor at the REST boundary, `env > server > default` precedence, and a hook-level test where a recall slower than the timeout aborts while a wider one lets it through.

Verified end-to-end against a real server binary (not mocks), since a handshake-settings bug is exactly the kind that mock suites structurally cannot catch:

```
1. baseline handshake      -> request_timeout_ms: 30000, source: default
2. PUT /v1/settings/defaults {"request_timeout_ms": 45000, "auto_save_min_events": 7} -> 200
3. handshake               -> request_timeout_ms: 45000, source: global
                              auto_save_min_events: 7,   source: global   # was 3/default before the fix
4. floor at the boundary   -> 0 => 400, 50 => 400
5. the real hook           -> ctx.timeoutMs = 45000 (was hardcoded 5000)
```

Gates: `test`, `lint` (0 issues), `test-hooks` (115), `test-client` (83), `docs-check`, `client-check`, ui build, openclaw (142) and pi (43) suites — all green.


eleboucher:
> _Originally by **eleboucher** on 2026-07-15 08:15 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/121#issuecomment-12597)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #122), which is the source of truth. Further review and merging happen there.

PR #49, opened 2026-07-15
Make the character limits honest, and remove the embedding ceiling instead of moving it

perfectra1n:
## Why

This started as a question: "do we really limit memories to 2000 characters?"

No. Nothing rejects a memory for being too long, and `content` is unbounded `text` in both drivers. But auditing every character cap turned up a pattern worth fixing: **every cap that was configurable was cosmetic, and every cap that was hardcoded silently degraded recall.**

The one that bites is `embedMaxItemChars = 8000`. It bounds what a memory's vector can *represent*, not what it stores. Text past it is kept and returned whole by `memory_get`, but vector recall cannot match it. The code already knew, in a comment: *"The stored vector represents only this prefix; recall won't match content beyond it."* It logged that at DEBUG, and had no env override. So the ceiling was unreachable, undiscoverable, and pointless on a backend with a larger context window. It is the most likely answer to "why can't memini find this?"

Four things came out of that audit, in dependency order: fix the length gates that were quietly wrong, make the budgets tunable, stop the clients truncating in secret, and then remove the embedding ceiling instead of moving it.

## What changed, by commit

**fix(extract): count runes, not bytes.** Four length gates compared bytes while meaning characters. UTF-8 spends up to three bytes per CJK character, so non-ASCII prose hit them at a third of the nominal length, silently and in both directions. The ceilings were too strict: content that should have earned a durable tier fell back to the short-lived `working` tier and expired, and a memory recalled often enough to earn promotion was never promoted. The 20-rune floor was too lax the other way: `"bug not working 是的"` is 22 bytes but 18 runes, so it scored two markers and became a durable fact, while the same length of ASCII was correctly rejected. Every display cap in the codebase already counts runes and says so; these four were the outliers. ASCII behaviour is pinned unchanged on both sides of every boundary.

**feat(config): five budgets become settings.** `MEMINI_EMBED_MAX_ITEM_CHARS`, `MEMINI_RERANK_MAX_DOC_CHARS`, `MEMINI_RERANK_LLM_MAX_DOC_CHARS`, `MEMINI_CLASSIFY_MAX_CHARS`, `MEMINI_PROMOTE_WHOLE_MAX_CHARS`. Defaults reproduce the constants they replaced exactly, pinned by a test, because an `envDefault` is a struct tag and cannot reference a constant, so nothing else keeps the two equal. The first two were in `deprecatedVars` as "now a fixed internal default"; those entries had to go, since gendocs renders that list into the generated reference's Removed table and would otherwise publish the same variable as both live and removed. `upgrading.md` said it by hand and is corrected too. `cmd/bench` and `cmd/qa` now read the same exported defaults rather than copied literals, because a benchmark against a stale budget measures a deployment nobody runs.

**feat(capture): the server sets the turn-capture bounds.** Every integration cut a captured turn at a hardcoded 1000 user chars and 3000 assistant chars, client-side, unmarked. How much of a turn is worth keeping depends on the store and the recall budget, which the server knows and the hook does not, so they become ClientSettings resolved in the handshake, beside `min_capture_chars`. Three silent bugs went with the literals: JS `.slice()` indexes UTF-16 code units and could split an emoji into a lone surrogate, putting invalid UTF-8 on the wire; `slice(0, 0)` is `""`, so the new 0-means-uncapped value would have captured nothing; and a non-numeric bound replaced the whole turn with just the marker. hermes also cut the memory *body* at 4000 chars on direct writes, which had no counterpart anywhere, so it is gone.

**feat(chunk): the splitter.** Pure functions, no callers. Cuts prefer a paragraph, then a line, then a sentence, then a word, searching each separator backwards across the window. Backwards is what makes it pack instead of fragment: the last boundary before the limit uses the most budget. An 11k-rune document becomes 12 chunks averaging 1100 of the 1200 allowed, none mid-sentence. Content at or under `MEMINI_CHUNK_MIN_CONTENT` produces *no* chunks, which is what makes "nothing changes for short memories" structural rather than a tuning coincidence.

**refactor(postgres): filterClause can qualify its columns.** A no-op today. `memory_chunks` carries `namespace` and `embedding` under the same names as `memories`, so the moment a query joins them every unqualified column is ambiguous and Postgres rejects the statement. sqlitevec's equivalent has always been aliased.

**feat(store): the capability, on both backends.** `store.ChunkStore`, `Memory.Chunks`, and ten conformance subtests both drivers pass identically. The backends reach chunks very differently, which is exactly why the shared suite matters: Postgres uses a child table with `ON DELETE CASCADE` and gets three cascades free, while vec0's only key is its rowid and it has no metadata-filter plan, so its chunk vectors live in `vec_chunks` keyed by rowid alone with a mapping table beside them and every cascade written by hand. `Reassign` is the one path a foreign key cannot cover, and both need it.

**feat(recall): the union, behind `MEMINI_CHUNK_EMBED`.** Recall's vector leg becomes `VectorSearch` union `ChunkVectorSearch`, merged per memory at its best score. Chunks are built by the backfill loop rather than at write time.

**fix(rerank): judge the matched chunk, not the memory's prefix.** Chunking and reranking fought each other, and rerank won. Recall handed the reranker the whole memory, and both backends then cut it to their own budget (300 bytes for the LLM reranker, 2048 runes for the cross-encoder), so a memory surfaced because of a passage deep in its content was judged on a prefix that did not contain it. Not a demotion: the rerank pool is deliberately deeper than `k`, and `rerank.Reranker`'s contract is explicit that an omitted candidate is DROPPED. Chunking paid the embedder to find the memory and the next stage discarded it. `memory_chunks` now stores each chunk's text, `ChunkVectorSearch` carries the winner out on `store.Scored.MatchedChunk`, and recall judges that.

**feat(chunk): the score weight, the gauge, and the benchmark.** `MEMINI_CHUNK_SCORE_WEIGHT` is the knob for max-pool's length bias, and it was briefly unreachable: the field, the option and the multiply existed with no config field and no wiring, so the remedy was absent while the bias was live. The chunk backfill gained the pending gauge its embed-backfill model already had. And `cmd/bench` gained a `-chunk` mode, which earned its place immediately (below).

- **Chunking without the rerank fix would have been worse than useless for its target user.** Both flags default off, so nothing shipped broken, but the combination is precisely what a long-memory operator reaches for, and it is the same bug as the original one layer up: a stage reads a prefix and silently concludes the wrong thing.
- **Adding a field to `store.Scored` was the interesting part of that fix.** Four sites in `internal/search` rebuilt a `Scored` from scratch to change its score, so `MatchedChunk` was silently dropped between retrieval and rerank and the fix did nothing until they were found. The review round below made the rule structural: adjusters now go through `Scored.WithScore`, and fusion backfills the field from whichever leg carries it, so survival depends neither on convention nor on which leg is fused first. This is exactly the coupling the design avoided by freezing `Scored`, and reopening it cost what it predicted.
- **Chunk text is stored, not recomputed.** Recomputing from content would be exact only while the split config is unchanged, and wrong the moment it is not.
- **A chunk-size change is not retroactive**, and the config doc says so. The backfill looks for memories with no chunks, so re-tuning reaches new memories only; existing chunks stay valid at the old granularity. Making it retroactive needs the split recorded per row plus a rebuild path, which is a schema change that belongs in its own review.
- **The benchmark found a real bug, so it is worth reading that bullet twice.** With chunking on and *zero* memories long enough to chunk, the sample's injected-token count moved from 78 to 81. `mergeVectorLegs` re-sorts, and its ID tie-break is not the store's order for equally-scoring rows, so merging an empty chunk leg reordered tied hits: a feature that did nothing still changed recall. An empty chunk leg now returns the store's rows untouched, the numbers match exactly, and a test pins it. "Purely additive" is only true if the path with nothing to add is bit-for-bit the path that existed before.
- **`ChunkVectorSearch` is a separate method, not a change to `VectorSearch`, and this is the load-bearing decision.** `VectorSearch` has six callers and three of them destroy data: write-dedup's coalesce/supersede, contradiction routing, and the maintenance dedup sweep. Chunk similarity is max-pooled, which makes a long memory a near-duplicate of anything matching any one of its paragraphs. Pooling inside `VectorSearch` would have had those three tombstone unrelated memories. My own plan said "max-pool VectorSearch" until this surfaced. Only recall reads the union, so turning chunking on can add hits and raise a score, and can never remove, tombstone, or invalidate anything.
- **Chunks are built in the background, so a long memory is fully searchable shortly after it is written, not instantly.** A long memory is many sequential embedder round-trips; doing that inside `Remember` would blow `MEMINI_WRITE_EMBED_TIMEOUT` for precisely the writes this feature helps, degrading them to `pending_embed`. The document vector works throughout, so the gap is in extra recall, not recall.
- **Discovery is a query (`ListUnchunked`), not a `pending_chunks` metadata flag.** Rows predating chunking carry no flag to find them by, so a flag would mean rewriting every row before the feature could do anything. It also makes every chunk-dropping path self-heal: a content rewrite or `memini reembed`'s model swap clears the rows, and the query finds them again next tick. The plan called for the stamp plus an audit of every Upsert caller; neither turned out to be needed.
- **`reembed` drops chunks on purpose, and now says so with an explicit empty slice.** The model just changed, so old chunks live in a space incomparable to the new document vector and keeping them would leave the index silently mixing two models. It is the one caller whose chunks go stale without a content change, which is exactly what the contract's empty-slice form exists for. The call site says so, because the tempting "fix" is the actual bug.
- **`MEMINI_CLASSIFY_MAX_CHARS` is refused between 1 and 19.** A ceiling under the extractor's 20-rune floor admits nothing, so it reads as a tight bound and acts as an off switch. `0` stays legal because it says "off" plainly.
- **`MEMINI_RERANK_MAX_DOC_CHARS=0` does not disable truncation**, it falls back to `MEMINI_RERANK_MAX_BATCH_CHARS`. I documented the opposite first; the CrossEncoder field's own doc had claimed it for longer.
- **Four copies of the client truncation survive**, because the standalone plugins ship with no build step and cannot import `@memini/client`. They are held together by `capture-vectors.json`, replayed by all four, in the shape `derivation-vectors.json` already uses. They needed it: the client core and the opencode copy disagreed on a NaN bound from birth, each passing its own tests.
- **The display caps stay fixed** (`conciseContentMax`, `childTitleMax`, `eventSummaryMax`, `defaultListLimit`). They cannot hurt recall, so a knob is permanent cost for nothing, and `deprecatedVars` shows that call was already made.
- **`lengthBonus` still counts bytes.** It is the same inconsistency as the rune fix, but it is a confidence-scoring weight rather than a cliff, so changing it moves what classifies and wants bench data first.
- **Emdashes in new code comments are kept**, since `origin/main` has 1124 of them across 178 Go files. Commit and PR prose has none.

## Round three: reviewing the whole branch, then fixing what it found

A third review pass ran over the complete diff: eleven finder angles, one adversarial verifier per surviving candidate, then a gap sweep. Fifteen findings survived verification and all fifteen are fixed, in the nine commits from `fix(chunk): floor the boundary backscan` onward. The four worth reading in full:

- **The backfill destroyed the document vector of every memory it chunked.** Its Get-then-Upsert round-trip could never carry the vector (Get does not load it, and both drivers read an empty Embedding as clear-the-vector), and nothing stamped `pending_embed`, so there was no repair path: enabling chunking silently removed every long memory from the document leg, write-dedup, and contradiction routing, and disabling it did not bring them back. Invisible to the whole test suite, because every post-backfill assertion passed through the chunk or keyword leg. The fix is structural rather than a patch: `ChunkStore.PutChunks` writes chunk rows only, under an updated_at guard inside the store's own transaction, which also closes the lost-update window that let the old code silently revert a concurrent `Reinforce`. A conformance case now asserts the document vector survives the backfill on both drivers.
- **The server-resolved capture caps never reached a client.** The REST layer maps ClientSettings by hand, and the two new fields were not in the mappers, so the handshake omitted them and a settings PUT of either was silently dropped. Every client fell back to catalog defaults equal to the old hardcoded values, which is why all tests stayed green around a dead feature. Fixed, together with `auto_save_min_events` (the same gap, from an earlier change), and a reflection round-trip test now fails by field name when a future field skips the wire. Fixing it would also have activated a latent bug in opencode, the one integration where a server value beat an explicit env var, so that precedence is fixed in the same round.
- **The TTL sweeper leaked chunk rows on sqlite, and the conformance suite was structurally blind to it.** `DeleteIfExpiredBefore` was the one delete path without the chunk cascade (the call meant for it had been pasted into `Delete` twice), so orphaned rows and vectors accumulated without bound and kept consuming KNN candidate slots. The expiry-cascade test passed anyway: `assertNoChunks` probed through `ChunkVectorSearch`, whose join back to `memories` is precisely what an orphan can never satisfy. The sweep now cascades, a startup migration purges orphans an earlier binary left, and the assertion counts rows through the new `CountChunks`, which a broken join cannot fool.
- **The splitter wasted half its budget on suffixes.** After a paragraph-boundary cut, the overlap rewind let the backscan re-pick the separator it had just consumed, emitting a chunk that was a pure suffix of its predecessor. A 46k-rune document hit `MaxChunks` at 78% coverage with 32 of its 64 chunks redundant and the tail unsearchable. The backscan now floors at the previous cut, so every chunk must carry new text.

The rest, one line each. `Memory.Chunks` becomes preserve-or-clear by content fingerprint, so promotion stamps, link-only consolidation, and metadata-only updates stop wiping a long memory's chunk index and re-paying its multi-chunk embed (the old always-clear contract also pointed callers at a `pending_chunks` flag that no code ever implemented, and following that comment would have caused the exact staleness it warned about). The backfill queue pages with a cursor and judges failures by run length, so one poison row the embedder always rejects, or a run of rows the splitter declines, can no longer wedge it (the old order was deterministic, so on sqlite that wedge was permanent). The pending gauge reports the queue's true depth instead of a number computed from the batch, which could never exceed 25 and read 0 through a ten-thousand-row backlog. `MEMINI_CHUNK_MIN_CONTENT` below the chunk size is refused at boot instead of silently turning every mid-sized memory into a whole-content chunk that duplicates its own document vector. The chunk KNN budget now composes the pooling and filter over-fetches its comment only claimed to, on both backends. `ListUnchunked`'s per-minute scan gets an expression index on content length in both schemas. The bench drain fails when rows remain unchunked instead of measuring a half-built index with a green exit. And the embed truncation log is WARN, which the upgrade guide had been promising it already was.

## Verification

- `go build` / `go vet` / `golangci-lint run` / `go test ./...` green, plus `test-hooks`, `test-client`, `docs-check` and `client-check` (both regenerate idempotently).
- `go test -tags=integration ./...` green against a fresh `vchord-postgres:pg18-v1.1.1` container. The thirteen chunk subtests pass identically on sqlite and on real Postgres, and the chunk suite now also passes twice in a row against the same reused database (the subtests that create rows they never chunk state their own preconditions, as `Reassign` already did).
- All six integrations green: client core 98, Claude Code hooks 113, pi 44, openclaw 70, opencode 82, hermes 60, openwebui 37 (skipped locally without pydantic, so its helper was checked directly against the same 12 vectors).
- **The actual bug, end to end.** A test buries a phrase past the embed cap, wrapping the fake embedder in `Batched` exactly as `cmd/memini` wires it. Without chunking the document vector cannot reach the phrase; after `BackfillChunks` the chunk leg does. It asserts on the chunk leg rather than on `Recall` deliberately: `Recall` fuses keyword with vector and keyword finds the phrase either way, which is exactly why this bug is so quiet. I checked, and a `Recall`-based assertion passed with the backfill removed, so it would have proved nothing.
- **Benchmark, both ways.** `go run ./cmd/bench` and `./cmd/bench -chunk` now produce identical recall, MRR, injected tokens and token-efficiency (78 tokens, 50.565%; only latency differs, which is noise). The run is deterministic, which is how the tie-break bug above was visible at all. One honest limitation: no bench dataset has a memory long enough to chunk (the longest item across all three is 368 runes against a 1200-rune floor), so this run confirms that short memories are untouched, on real data rather than by assertion, but it cannot yet settle `MEMINI_CHUNK_SCORE_WEIGHT`. Measuring the length bias needs a long-document dataset, which this PR does not add.
- **The chunk/rerank interaction is pinned by a test that models the reranker faithfully** (it truncates its candidate to 300 bytes, as `rerank/llm.go` does, and omits what it cannot see the evidence in). It fails without the fix in both directions: I reverted only the candidate change and it went red.
- Every fix in this PR was checked by breaking it again: the rune gates, the defaults-drift test (a typo'd `800` for `8000` reports precisely that), the capture vectors (reintroducing the original NaN bug turns hermes red), and the chunk end-to-end test above.
- The round-three fixes carry their own regression tests, aimed at the blind spots that let the bugs through: the document vector asserted through the DOCUMENT leg after a backfill (every prior assertion used the chunk or keyword leg), a gauge fixture with a queue deeper than one batch (the only shape that can catch a batch-capped gauge), the poison-row and outage fixtures, a degradation test that runs against a store genuinely lacking the capability (its predecessor ran against sqlitevec, which has it, so its branch was never taken), and MatchedChunk survival pinned through both fusion orders, RRF, and the temporal path that production wires at boost 0.40. The bench parity run still reads 78 tokens and 50.565% with `-chunk` on and off.


eleboucher:
> _Originally by **eleboucher** on 2026-07-16 07:20 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/124#issuecomment-12688)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #129), which is the source of truth. Further review and merging happen there.

PR #50, opened 2026-07-16
Give REST an update-in-place route, and stop re-embedding content that did not change

perfectra1n:
## Why

This started as a claim made mid-task, while scripting an edit across 88 memories:

> PATCH/PUT are both 405, there's no REST update-in-place route (only GET/DELETE on {id}, POST only on the collection which is create/upsert and would force a re-embed). The correct in-place tool is MCP memory_update (proven, no re-embed).

The 405s are real. Almost nothing else in that is.

There *is* a REST update-in-place route: `POST /v1/memories` binds an optional body `id` and upserts, which `RememberInput.ID` and the spec both say plainly. And `memory_update` was never re-embed-free. It is `svc.Get`, overlay fields, `svc.Remember` with the same ID, which is the identical service call `POST` with an `id` lands on. There is no `service.Update` at all. Both re-embedded, unconditionally, so the 88 memories would have cost 88 embedder calls either way.

The conclusion was wrong and the instinct was right, for an unstated reason. `memory_update` really is the correct tool, because of its merge semantics, not its embedding behaviour: `POST` with an `id` is a whole-record replace that drops every metadata key the caller did not resend. 88 partial POSTs would have silently destroyed data. The 405 prevented that by accident.

So two real gaps, and they are the same gap seen from two sides. **The re-embed is genuinely wasteful**: a tags-only edit pays a full embedder call on byte-identical content, and `Remember` embeds *before* it loads the existing row, so it could not have compared them even in principle. **And the surfaces have drifted**: every partial-update rule lives in the MCP handler, REST has no equivalent, and the route it pushes callers toward is the one that loses their data.

## What changed, by commit

**feat(store): add GetEmbedding so a write can preserve a vector it is not recomputing.** The obvious implementation of "reuse the vector when content did not change" is a silent data-loss bug, and this is the finding that shaped the whole change. `existing.Embedding` is *always* empty: `Memory.Embedding` is documented as "required when writing to the store and omitted from API responses", `memoryColumns` has no embedding column, and Postgres says so on `List`. `Upsert` then reads an empty `Embedding` as "drop the vector". So `Get` a memory, hand it back with the embedding untouched, and the row silently leaves the vector index. `Upsert`'s contract is right and stays, because a degraded update *must* drop the vector rather than leave a stale one pointing at replaced content, so a write that wants to keep a vector it is not recomputing has to read it back explicitly. This is the same trap #49 hit in the chunk backfill, from the other direction: that one had to stop writing the vector, this one has to start reading it.

**perf(service): reuse the stored vector when an update leaves content unchanged.** The existing-row load moves above the embed and the embed becomes conditional. Content is compared after `scrubInput` and `sanitizeContent` have rewritten it, so it is the exact string the embedder would have received, which makes the reused vector identical to a recomputed one rather than close to it. A create skips the load entirely (`in.ID` is empty), so the fresh-write path pays nothing and `POST` with an `id` gets the saving for free. The has-a-vector check is load-bearing rather than defensive: a row left vectorless by a degraded write has matching content and nothing to reuse, and it is exactly the documented recovery path, so it must still embed for real.

**refactor(service): extract Update, the partial-edit path both surfaces share.** `Get` plus `Remember` with the current ID, so an edit still runs validation, redaction and sanitization exactly as a fresh write does, and dedup/consolidation/contradiction routing stay fresh-write-only. Fields become pointers, because the `""` sentinel could not distinguish "omit to keep" from "set to empty" and a summary could never be cleared. Three things it fixes rather than carries over, since sharing them with REST would have made them worse: tier validation and the value-gate drop now return `ErrInvalidInput`, because `statusFor` maps anything unrecognized to **500** and the bare `fmt.Errorf` would have reported a caller's typo as a server fault; and `mergeMetadata` allocates, because `Remember` mutates the map it is handed and was being given the one off the stored record.

**refactor(mcp): repoint memory_update onto service.Update.** The handler becomes a translation layer. The wire schema is unchanged: `*float64` already rendered as a plain number, so `*string` renders as a plain string, and the only regenerated `mcp-tools.md` diff is the metadata description. `TestUpdateToolContentOnly` passes unmodified, which is the check that the extraction preserved the semantics rather than approximated them.

**feat(api): add PATCH /v1/memories/{id}.** Spec-first, so the route is generated. The generated request type is all optional pointers, mapping onto `UpdateInput` one to one, and keeping `"metadata": null` (no change) distinct from `{"metadata": {"k": null}}` (delete `k`) for free.

## Behaviour changes worth a changelog line

All three are on inputs that are meaningless today, which is what makes them worth taking.

- `{"summary": ""}` cleared nothing and now clears the summary. This is the bug the pointers fix.
- `{"content": ""}` was a silent no-op and now errors. Silently ignoring a caller who sent empty content was the worst of the three options.
- `{"metadata": {"k": null}}` stored a null and now deletes the key.

- **The skip lives in `Remember`, not in `Update`, and that is the load-bearing call.** Putting it in `Update` would compare pre-scrub content, correct only because `redact.Secrets` and `sanitize.Clean` happen to be idempotent on already-processed text, which is fragile reasoning to bake into a correctness check. It would also have missed the `POST`-upsert win. The alternative of passing a reusable vector through `RememberInput` was rejected outright: it puts a raw `[]float32` on the most-used struct in the service, where any caller can pass a vector that does not match its content.
- **`GetEmbedding` is required on the interface, not optional.** Reading back a vector is core to a vector store, `internal/store` has no out-of-tree implementers, and getting it wrong loses data silently, so the conformance suite should enforce it on every driver. `EmbedModelStore` is optional because recording a model name is genuinely driver-specific; this is not.
- **`(nil, nil)` for a vectorless row is the whole point of the method.** The caller must tell "nothing to reuse, go embed" apart from "no such memory". sqlitevec gets there with a LEFT JOIN for exactly the reason `Reassign`'s is one, and Postgres scans into a `*pgvector.Vector` so NULL lands as nil instead of failing the scan.
- **A `GetEmbedding` error falls through to embedding rather than failing the write.** A redundant embed beats failing a write that would otherwise succeed.
- **`importance: 0` stays broken, and is documented rather than fixed.** `resolveImportance` uses a `!= 0` sentinel on a `float64`, so an explicit zero is ignored today via `memory_update` too. Making it honest means pointer-izing `RememberInput.Importance` across `rest.go`, `promote.go`, `consolidate.go`, the importer and maintenance, which is its own change. The PATCH docs do not claim it works.
- **PATCH merges metadata, POST with an `id` replaces it.** Keeping both is deliberate: replace is what an upsert means. The divergence is documented on both, pointing at each other, in the spec and in the Go docs, because it is the detail most likely to be assumed rather than read.
- **A null metadata value deletes the key**, following RFC 7386. There is currently no way to delete a metadata key at all; the only alternative is a wholesale re-POST of every other key, which is both racy and lossy.
- **`Update` reads the row twice** (`s.Get`, then `Remember`'s own load). That is true today via the MCP path as well, so it is not a regression, and collapsing it means teaching `Remember` to accept a preloaded row. Left alone.
- **`s.Get`, not `s.store.Get`.** `s.Get` logs the activity event the MCP path has always emitted. Switching would have dropped it silently.

## Verification

Full unit suite and `golangci-lint` clean. The full integration suite passes against a real vchord Postgres, including the new conformance case on both drivers.

The tests that matter here were checked by mutation, because three of them assert things a passing test can trivially fake:

- Reusing `existing.Embedding` (the naive implementation) fails the reuse test.
- Dropping the has-a-vector check strands degraded rows, and fails the recovery test.
- Swapping `invalidInputf` for `fmt.Errorf` turns the PATCH bad-tier case into a **500**, which also confirms the generated binder does *not* validate the tier enum, so that guard is load-bearing rather than belt-and-braces.

The reuse test asserts the memory is **still returned by `Recall`**, not merely that the embedder went uncalled. The vector-wipe bug spends zero embedder calls too, so a count-only assertion passes straight through it while the corpus rots.

Worth noting what the existing suite could not do: `TestRememberUpdatePreservesPendingEmbed` and `TestRememberUpdateRecoveryClearsPendingEmbed` both change content, so both bypass the new branch entirely. They pass, and they prove nothing about it. That is why the recovery case had to be written rather than assumed.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -547,6 +547,35 @@ paths:
             application/json:
               schema: { $ref: "#/components/schemas/Memory" }
         "404": { $ref: "#/components/responses/Error" }
+    patch:
+      operationId: updateMemory
+      tags: [memory]
+      summary: Update a memory in place
+      description: >
+        Changes only the fields present in the body; anything omitted carries
+        over from the stored record. metadata merges into the existing metadata
+        key-by-key and an explicit null deletes that key — POST /v1/memories
+        with an id replaces metadata wholesale instead.
+
+
+        The write-time lifecycle still runs (validation, secret redaction,
+        content sanitization), so this is not a bare field write. Content is
+        re-embedded only when it actually changes, so a tags- or metadata-only
+        update costs no embedder call.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema: { $ref: "#/components/schemas/UpdateMemoryRequest" }
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema: { $ref: "#/components/schemas/Memory" }
+        "400": { $ref: "#/components/responses/Error" }
+        "404": { $ref: "#/components/responses/Error" }
+        "500": { $ref: "#/components/responses/Error" }
     delete:
       operationId: forgetMemory
       tags: [memory]
@@ -1069,7 +1098,12 @@ components:
             (the default) — it only raises to semantic/procedural.
         summary: { type: string }
         tags: { type: array, items: { type: string } }
-        metadata: { type: object, additionalProperties: true }
+        metadata:
+          type: object
+          additionalProperties: true
+          description: >
+            Replaces the stored metadata wholesale when this write carries an id
+            (an upsert). Use PATCH /v1/memories/{id} to merge key-by-key instead.
         importance: { type: number, format: double, minimum: 0, maximum: 1 }
         level:
           allOf:
@@ -1115,6 +1149,52 @@ components:
             request namespace "acme/phoenix/api"), and the write lands there
             instead. Ignored for non-durable writes: an episodic/working memory
             always stays in the request namespace regardless of visibility.
+    UpdateMemoryRequest:
+      type: object
+      description: >
+        A partial edit: every property is optional and anything omitted keeps
+        its stored value. Sending an explicit value writes it, so summary: ""
+        clears the summary rather than being ignored.
+      properties:
+        content: { type: string }
+        summary: { type: string }
+        tier:
+          allOf:
+            - $ref: "#/components/schemas/Tier"
+          description: Move the memory to this tier. Omit to keep the stored tier.
+        level:
+          allOf:
+            - $ref: "#/components/schemas/Level"
+          description: Relabel derivation provenance. Omit to keep the stored level.
+        tags:
+          type: array
+          items: { type: string }
+          description: >
+            Replaces the stored tag set wholesale. Omit to keep it; send an
+            empty array to clear it.
+        metadata:
+          type: object
+          additionalProperties: true
+          description: >
+            Merges into the stored metadata key-by-key: listed keys are added or
+            overwritten, unlisted keys are untouched, and a key sent with a null
+            value is deleted. POST /v1/memories with an id replaces metadata
+            wholesale instead.
+        importance:
+          type: number
+          format: double
+          minimum: 0
+          maximum: 1
+          description: >
+            Omit to keep the stored importance. Note that 0 is currently
+            indistinguishable from omitted and keeps the stored value.
+        confidence:
+          type: number
+          format: double
+          minimum: 0
+          maximum: 1
+          description: >
+            Omit to keep the stored confidence; ignored for short-term tiers.
     SupersedeRequest:
       type: object
       required: [by]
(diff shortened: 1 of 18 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-07-16 07:20 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/125#issuecomment-12689)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #128), which is the source of truth. Further review and merging happen there.

PR #51, opened 2026-07-16
Make degraded writes visible: pending-embed status on stats, doctor, healthz, and the UI

perfectra1n:
## Why

When the embedder is unreachable and a write-embed timeout is configured, a write is stored vectorless with `metadata.pending_embed="true"` and waits for the backfill loop to re-embed it. Until that happens the memory matches keyword retrieval only. Nothing showed this to a human: the flag was buried in the drawer's raw metadata JSON, the backlog gauge went only to Prometheus, and `/healthz?verbose=1` said nothing about the reranker.

The symptom that started this was memories in the UI that looked like they held placeholder values, with no way to tell whether they were still queued for embedding or reranking. Auditing that suspicion showed most of those values are real: confidence 0.40 is the fresh-fact seed (0.5626 is that seed corroborated three times), importance 0.6/0.3/0.1 are the tier seeds, and reranking is purely query-time, so there is no per-memory rerank state to wait on. So this PR does two things: it surfaces the one genuine "still queued" state at every level a human looks, and it labels the seeded defaults so they stop reading as placeholders.

## What changed

**Stats.** `service.Stats`/`StatsAll` count live memories carrying the flag, exposed as a required `pending_embed` integer on `GET /v1/stats` (spec, generated Go/TS, and docs regenerated). Counting the flag rather than "rows without vectors" is deliberate: it counts exactly what the backfill will process; unflagged vectorless rows are an fsck concern.

**Doctor.** The write-path signals section gains `pending embed (vectorless, awaiting backfill): N`, and the section now prints when the backlog is the only nonzero signal. The tally is gated to live rows so it matches stats and the backfill; without the gate, a flagged row that is later superseded or expires would be reported forever while never draining.

**Healthz.** `/healthz?verbose=1` gains a `deps.reranker` block (`configured`, `ok`, `last_error`, `last_success`) fed by a `recordingReranker` wrapper on the recall path, the same pattern as `recordingEmbedder`. `llmDepBlock` is renamed `optionalDepBlock` since two optional deps now share it; the wire shape of the existing blocks is unchanged. `storeDepBlock` now stamps `last_success` on a healthy ping, so consumers can tell "probed healthy just now" from "never called" without special-casing the store.

**UI.** Memory cards and the drawer show an amber `awaiting embed` chip keyed off the metadata flag (the backfill deletes the flag on success, so the chip self-clears). The Dashboard grows an "Awaiting embedding" tile that only renders when the backlog is nonzero. The Health view gets a read-only Pipeline panel above fsck/dedup: store, embedder, LLM, and reranker rows with status dots and last-error/last-success detail, plus the backlog count. The backlog row is tri-state: it shows "unavailable" rather than a reassuring "none" when the stats fetch itself fails. Confidence values are labeled `seed` at exactly 0.40 and `corroborated` above it, and importance values that equal their tier seed say so; the displayed numbers derive from the shared constant rather than being retyped in the copy.

**Flag contract.** The key and sentinel now live in one place, `memory.PendingEmbedKey`/`memory.PendingEmbedValue` with a `Memory.PendingEmbed()` predicate; the write stamp, backfill filter, stats count, doctor count, and the MCP degraded check all use it instead of each spelling out the literal.

- No per-memory rerank status exists or is invented: reranking reorders candidates per query and persists nothing, so the reranker gets liveness reporting only.
- The UI mirrors the Go seed constants (`seedImportance`, `ConfidenceSeedFresh`) with a keep-in-sync comment rather than a server-exposed provenance field. Those constants have never been retuned since introduction, the feature is display-only, and since the copy derives from the constant a future retune can only miss a label, never show a wrong number.
- The verbose-healthz response shape stays out of the OpenAPI spec (its 200 is schemaless there on purpose); the UI types it by hand next to the other hand-written interfaces, with a comment pointing at the Go source of truth.

## Verification

- `mise run test`, `lint`, and `docs-check` green; the oapi-codegen and UI-schema drift gates are clean; `pnpm run typecheck` and `pnpm run build` clean.
- Live end-to-end pass, not just suites: a server built from the branch, embedder pointed at a dead port with a small write-embed timeout and rerank configured against another dead port. A write landed flagged; `/v1/stats` showed `pending_embed:1`, verbose healthz showed `embedder.ok:false` with the error and `reranker.configured:true`, doctor printed the backlog line, and a real browser showed the chip on the card and drawer, the seed and tier-default labels, the Dashboard tile, and the Pipeline panel with the store row healthy, the embedder row failing, and the backlog at 1. Pointing the embedder at a working stub and waiting one backfill tick cleared every surface, including the metadata flag itself.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -1477,6 +1477,7 @@ components:
           expired,
           superseded,
           low_confidence_durable,
+          pending_embed,
           total_accesses,
           avg_importance,
         ]
@@ -1495,6 +1496,9 @@ components:
         low_confidence_durable:
           type: integer
           description: Live durable memories whose decayed confidence is below the demote floor — reclaimable, uncorroborated debris.
+        pending_embed:
+          type: integer
+          description: Live memories saved without a vector because the embedder was unreachable at write time; the backfill loop re-embeds and unflags them, so a persistently nonzero value means the embedder is still down.
         total_accesses: { type: integer }
         avg_importance: { type: number, format: double }
         last_write_at: { type: string, format: date-time, nullable: true }
--- a/cmd/memini/depstatus.go
+++ b/cmd/memini/depstatus.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/eleboucher/memini/internal/embed"
 	"github.com/eleboucher/memini/internal/llm"
+	"github.com/eleboucher/memini/internal/rerank"
 	"github.com/eleboucher/memini/internal/server"
 )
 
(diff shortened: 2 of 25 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-07-16 07:20 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/126#issuecomment-12690)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #127), which is the source of truth. Further review and merging happen there.

PR #52, opened 2026-07-16
fix(plugin): session briefing rendered nothing after the T6 wire change; harden + freshen the briefing surface

perfectra1n:
## The headline: the Claude Code/Codex hook has been injecting nothing since the T6 wire change

Commit 2271aa1 (v0.7.x, T6) changed the REST briefing sections from flat `Memory` items to nested `BriefingItem{memory, from}`. The pi and openclaw integrations were updated to unwrap `b.memory` — but `plugin/scripts/session-start.mjs` was not. `formatMemory` read `m.summary || m.content` on the wrapper, every bullet rendered null, all blocks came out empty, and the `blocks.length === 0` early return exited **before the memory directive was emitted**. Net effect against a current server: the SessionStart hook injected nothing at all — no briefing, no save directive.

The hook's own tests masked this because their mock servers still served the old flat shape. This PR converts the fixtures to the real nested wire (keeping one flat fixture to pin back-compat with pre-T6 servers via `item?.memory ?? item`) — the regression test that was missing.

## What else is in here

Building on the fix, six small quality/hardening improvements to the briefing surface (each its own commit, each TDD'd):

- **`from` provenance renders** — the MCP server instructions tell the model to "read `from` … to learn where knowledge actually lives", but the hook never showed it. Bullets now carry ` (from <ns>)` when inherited (ancestor/personal/link).
- **Anti-double-briefing note** — the injected block's comment now says it replaces a `memory_briefing` call, so hook clients don't redundantly re-call the tool the server instructions tell them to call at session start. A reachable-but-empty namespace now says `(no stored memories yet for this project)` (an unreachable server still makes no emptiness claim).
- **Honest truncation** — bullets cap at 280 code points with a `…` marker only when actually truncated (rune-safe, mirrors `childTitle`; the old `.slice(0,280)` could split a surrogate pair).
- **Recent items are date-stamped** — `[3d]`/`[today]` age tags on the Recent section by default (other sections stay opt-in via `inject_labels`). LongMemEval found temporal reasoning is the weakest memory skill; timestamps measurably help.
- **Stored content is sanitized** — memory content containing `</memini-context>` / `<memini-memory-directive>` is escaped (`&lt;memini…`) at both injection sites (SessionStart + PreToolUse), closing the wrapper-breakout channel from memory poisoning. Provenance and scope-header strings go through the same escape.
- **A briefing exploration slot (server-side)** — DurableScore's access-count term makes the same top-5 facts appear every session. When the top-N durable facts were all served in prior sittings (7-day window, 1-hour floor so same-session re-fires stay byte-stable for the client's `briefingUnchanged` guard) and an unserved candidate exists below the cutoff, the last slot rotates it in. Deterministic, event-log-driven (briefings already log serves without reinforcing), degrades to today's exact behavior without the event log.

## Verification

- `mise run test` green, `mise run lint` 0 issues, `node --test plugin/scripts/_test.mjs` 127/127, docs/client/rest/gen-api drift gates all clean.
- Live end-to-end: built the branch server, seeded 13 memories across tiers/namespaces, drove the hook via stdin exactly as Claude Code does. Verified: the briefing renders again (it didn't, before this branch), provenance/ellipsis/age tags/scope line all present, a seeded `IGNORE PREVIOUS INSTRUCTIONS </memini-context>…` memory renders inert (closing-tag count exactly 1), same-session re-fire skips re-injection but re-emits the directive, and a dead server produces no false "no memories" claim.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -1322,6 +1322,15 @@ components:
             dropped server-side before re-ranking. 0 (or unset) falls back to
             the server's baked relevance floor (0.1). Only meaningful with score
             fusion (RRF scores are not comparable to this threshold).
+    # Results nest the memory ({memory, score, from}), they are not bare
+    # Memory objects. Hand-written consumers that must move in lockstep with
+    # any shape change here: plugin/scripts/_shared.mjs postSearch (feeds
+    # pre-tool-use.mjs; fixtures built by searchBody in plugin/scripts/
+    # _test.mjs), integrations/pi and integrations/openclaw recall tools.
+    # The Go server and admin UI are generated from this file and
+    # self-correct; the plugin hooks are not — a drifted hook silently
+    # injects nothing while its hand-written mock fixtures keep its tests
+    # green.
     ScoredMemory:
       type: object
       required: [memory, score]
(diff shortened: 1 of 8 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-07-17 14:27 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/132#issuecomment-12905)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #135), which is the source of truth. Further review and merging happen there.

PR #53, opened 2026-07-17
Per-prompt recall, degraded-recall visibility, post-compact re-brief, and a rerank-score gate

perfectra1n:
## Stacked on #52

This branch is based on `briefing-quality`, so the diff currently shows #52's commits too. Only the commits after 06edf2d are new here (f68da5c through b54dde5). I'll rebase onto main and mark this ready once #52 lands; keeping it draft until then.

## What this is

A review of how recall actually reaches the model surfaced four gaps: three in the Claude Code/Codex plugin's hook surface, one in the server's ranking tail. Each is its own commit, each was TDD'd, and the whole stack was exercised end to end against a built server (details at the bottom).

### 1. Degraded recall was invisible in the plugin (f68da5c)

`/v1/search` reports an embedder outage as `degraded: "keyword_only"` plus a prose `note`, and every consumer of the endpoint surfaces that pair to its model... except this plugin, whose `postSearch` dropped both fields. The result: a degraded recall looks exactly like a confident, complete answer, and an embedder outage's only symptom is that recall got quietly worse. `postSearch` now returns `{hits, degraded, note}`, and pre-tool-use renders one escaped `[memini: ...]` warning line per injection block. The note renders only when hits exist, so a dead embedder doesn't stamp a bare warning onto every tool call, and only once per block rather than once per file.

### 2. Per-prompt recall: the prompt is the query (2d385d1)

Recall in this plugin was only ever file-path shaped: PreToolUse searches with `"<Tool> on <path>"`, which can surface memories about which files are touched but never about what the user is trying to do. Meanwhile the four knobs for prompt-driven recall (`recall`, `recall_limit`, `inject_recall_max_tok`, `inject_recall_min_score`) have existed in the shared client all along, and the opencode, pi, and openclaw integrations already consume them for their per-prompt recall. No Claude/Codex hook read them.

The new `UserPromptSubmit` hook searches with the user's words and injects the top hits, on the same guardrails as pretool: cache-only namespace resolution (degraded context skips silently and self-heals via Stop), fresh-turn echo filtering, wrapper-tag escaping on all recalled content, token budget, and the degraded warning from commit 1. Command-shaped prompts (`/`, `!`, `#` prefixes) and prompts too short to be a useful query skip the search entirely.

To keep consecutive related prompts from re-injecting the same memories, the hook keeps a per-session list of injected ids and excludes them server-side via `exclude_ids`, so the top hits are spent on context the conversation doesn't already carry. An older server that doesn't know `exclude_ids` 400s the whole request; the hook retries once without the field, degrading to "no server-side dedupe" instead of "no recall at all". The state dies with the context it describes (SessionStart, PreCompact, SessionEnd). Rider refactor: pretool's `formatHit` and fresh-turn filter moved into `_shared.mjs` now that this is their third consumer.

`source: "prompt"` is recorded on the activity event, so the Activity view distinguishes these recalls from pretool's with no UI change.

### 3. Post-compact re-brief, and MultiEdit parity (5322490)

The unchanged-briefing guard skips re-injection whenever the briefing bytes match the previous fire, and since #52 the server deliberately keeps briefings byte-stable within a sitting, so on `SessionStart` with `source: "compact"` the guard always fired. But compaction just summarized the injected briefing away with everything else. The guard's premise ("the identical block is already in context") is false on exactly this path, and its cost argument is moot because the rebuild already busted the prompt prefix cache. So the model came back from every compaction with a flush directive and no facts. Compact now falls through to a full re-injection and re-caches the same hash, so a later resume still skips; resume keeps the skip because its context is intact.

Also in this commit: `PostToolUse` has matched `MultiEdit` since the beginning, `PreToolUse` never did, so multi-file edits got buffered but never got a recall injection. Added to the hooks.json matcher and to the `inject_pretool_tools` default everywhere it lives (settings.ts, openapi.yaml, `DefaultClientSettings`), plus the generated mirrors.

### 4. `MEMINI_RERANK_MIN_SCORE`: a cross-encoder relevance gate (c8d7d92)

A reranker that only reorders still returns something for every query. On a query with no real answer, that something is the least-irrelevant noise, confidently ranked. The fused floor (`MEMINI_RECALL_MIN_SCORE`) can't cut this tail because min-max normalization inflates the best of a bad pool into competitive territory (the "poison" trade-off already noted in the recall code). Cross-encoder scores are different: they're calibrated absolute relevance, and `bench.RerankGateSweep` has measured that hit-vs-noise separation for a while. This puts it to work at runtime.

The gate drops candidates the cross-encoder scores below the threshold, across the whole `MEMINI_RERANK_POOL` before the k cap, so a gated-out top candidate still lets a below-cut survivor through. A fully-gated pool returns empty recall: the new `WithRerankEmptyVerdict` service option makes empty a verdict instead of the existing fall-back-to-composite-order path, which would otherwise undo the gate on exactly the queries it exists for. It's wired only when the gate is configured, so ungated deployments keep the fallback for backend pathologies; rerank failures (error, timeout) fall back either way, since a dead reranker never rendered a verdict. Gated-empty is recorded as `rerank_result{result="empty"}`, distinct from `ok` and `fallback`.

Guardrails: the single-candidate no-op shortcut in `CrossEncoder.Rerank` is disabled when gated (a `limit=1` recall must still be scorable to zero), and `MEMINI_RERANK_MIN_SCORE` with `MEMINI_RERANK=llm` refuses to boot, because the ordinal-list backend has no scores and a gate that silently never fires is worse than a loud misconfiguration. Default is 0 (off): thresholds are model-specific, so `docs/guides/tuning-recall.md` gained a section pointing at the bench sweep for picking one. The response `score` field still carries the fused score; rerank scores stay off the wire, so no consumer contract changes.

### 5. Cross-surface injection dedupe (ce0b22f)

The three injection surfaces (briefing, prompt recall, pretool recall) each kept private dedupe state, so one memory could enter the same context up to three times: briefed at startup, recalled for the prompt that mentions it, recalled again for the file it describes. They now share one per-session record of what the context already carries: an insertion-ordered map of injected memory id to content-identity hash.

Each surface consumes it the way its semantics allow. The prompt hook excludes recorded ids server-side (`exclude_ids`), spending its top-k on memories the conversation doesn't yet carry; that's id-based by necessity, since the server can't compare content it was told not to return. Pretool filters client-side and content-aware instead: a memory whose content changed since injection hashes differently and passes, so a `memory_update` mid-session still resurfaces (this deliberately does not send `exclude_ids`). The filter also accumulates across the files of a single pretool call, and the briefing records what it rendered so both hooks skip it.

The state's lifecycle mirrors the context itself: cleared on startup/clear/compact (SessionStart is now source-aware) and at PreCompact/SessionEnd, kept across a resume, whose context is intact. Everything rides the existing `inject_dedupe` knob, so `MEMINI_INJECT_DEDUPE=0` restores the prior always-inject behavior on every surface at once.

Two historical pretool pins change intentionally, and their tests now pin the new behavior: identical results for a *different* file no longer re-inject (the context already carries them; which file put them there is irrelevant), and the bounded-state test serves distinct memories per file since a shared one is deduped after its first injection.

### 6. Docs: the exact rerank-gate sweep invocation (4c278c3)

The gate section pointed threshold-picking at the Go symbol `bench.RerankGateSweep`; `cmd/bench` has carried the `-rerank-gate` flag for a while. The tuning guide now shows the runnable command, what the two columns mean, and how to read them.

### 7. The Claude plugin tracks memini MCP tool reads (43bfa3c)

The cross-surface dedupe covered the three hook surfaces but not the fourth way memories enter context: the model calling memory_recall / memory_briefing / memory_get itself. PostToolUse now matches the memini read tools (both MCP namings) and records the returned ids into the same state, via one depth-limited walker that covers all three response shapes. Recorded with a sentinel identity, because a concise tool response may truncate content: suppression for tool-sourced entries is by id alone, while hook-recorded entries keep their real hash and content-aware resurfacing. Reads never touch the event buffer; write tools stay untracked (their results carry no content, and the content came from the model).

### 8. hermes gets conversation-scoped injected-id dedupe (aee7153)

hermes was the one integration whose per-prompt recall kept no memory of what it had injected. An instance-scoped id set now backs prefetch exclusion (server-side exclude_ids with a retry-then-latch fallback for older servers that 400 the field, plus a client-side filter), recording from the recall/briefing tool handlers, and a deliberate bypass-and-reset in on_pre_compress, whose whole job is re-surfacing context that compression is about to summarize away.

### 9. openclaw's briefing tool feeds its injected-id state (b54dde5)

openclaw's memory_recall tool has always recorded what the model pulls; memory_briefing never did. The briefing executor now records every section item's id under the same session identity.

**Deliberately not changed:** pi's tool executors receive no session identity, so recording tool reads there risks cross-session contamination (the same hazard class the handshake cache work guarded against) — its auto-recall stream already dedupes against itself, and the tool-read tracking can follow if pi's host API grows a session handle in tool context. openwebui's filter is stateless by architecture and opencode exposes no memory read tools, so neither has anything applicable.

## Verification (follow-ups)

156/156 plugin tests, 66 hermes tests, 143 openclaw tests, all drift gates green. Live re-run of the same end-to-end driver: the prompt fired right after the briefing now injects nothing (previously it re-injected two just-briefed facts), `pre-compact` clears the state and the next prompt re-injects, pretool right after that re-injection filters to empty, and a fresh session with no state still injects with the degraded note.

## Verification

Per commit: `mise run test`, `mise run lint` (0 issues), `node --test plugin/scripts/_test.mjs` (145 tests), `mise run test-client` (102), and the docs/client/gen-api/rest/helm drift gates, all green.

Live end to end against a built branch server (sqlite, stub OpenAI-compatible embedder, stub Cohere-style `/rerank`), driving the hooks over stdin exactly as Claude Code does:

- Briefing trio behaves: startup injects, resume skips but re-emits the directive, compact re-injects the full briefing plus the recovery directive.
- First prompt injects the top hits; repeating the prompt surfaces only a not-yet-injected memory (`exclude_ids` at work); `pre-compact` clears the state and the next prompt re-injects.
- Slash and short prompts make zero server calls. MultiEdit gets a pretool injection.
- Killing the embedder mid-session produces the keyword-only warning in both hooks, and the server log shows the matching `embed_timeout` fallback. A byte-identical query served from the server's embed LRU is genuinely healthy and correctly carries no warning.
- With the gate at 0.9, an off-topic query returns zero results where ungated it returned noise normalized to score 1.0; an on-topic query returns exactly the relevant memory. `MEMINI_RERANK=llm` plus the gate refuses to boot with a pointed error.


eleboucher:
> _Originally by **eleboucher** on 2026-07-17 07:32 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/133#issuecomment-12855)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #134), which is the source of truth. Further review and merging happen there.

PR #54, opened 2026-07-18
Windowed injection cooldown and a per-file recall call gate

perfectra1n:
## Why

The cross-surface injection dedupe from #134 suppresses a re-injection forever within a conversation. That solved the "same three memories on every hook event" problem, but it left two gaps that show up clearly in production:

1. **No re-reminding.** Once a memory is injected, it never comes back for the rest of the conversation, even hundreds of tool calls later when the model's attention on that early context has decayed. Long sessions benefit from an occasional refresher; forever-suppression can't provide one.
2. **The recall call itself still fires on every event.** PreToolUse runs a full recall (embed, fuse, rerank) on every Edit/Read even when the client then suppresses the injection. A burst of edits to one file is a burst of identical recalls: wasted server work, wasted reranker calls, and an activity feed full of "served 3 memories" rows that were never actually injected.

Numbers from a 23 hour window of production activity events (1200 recall ops, 3638 served memories): 68% of serves re-served a memory that had already been served within the previous minute, 89% within 30 minutes, with a median re-serve gap of 14 seconds. The single most-served memory was returned 606 times in that window. A third of all recall ops served the exact same set as the op immediately before them, almost always an Edit/Edit/Edit burst on one file.

For prior art: Letta and Zep avoid this class of problem by replacing a memory block in place, which hook-based injection can't do (hooks only append). The closest ecosystem analogue is hermes-agent's Mem0 plugin, where per-turn retrieval into the prompt produced roughly 1% prefix cache hit rates; the mitigations discussed there (session-level prefetch, retrieval caching with a novelty gate) map onto memini's briefing and this PR's cooldown plus call gate.

## What

Injection dedupe becomes **windowed** instead of forever, and the pretool recall call gets a short per-file gate.

Three new `ClientSettings` knobs, following the existing knob machinery end to end (openapi, store, REST mappers, client `BEHAVIOR_KNOBS`, UI catalog, docs):

| knob | env | default | meaning |
|---|---|---|---|
| `inject_cooldown_ms` | `MEMINI_INJECT_COOLDOWN_MS` | 1800000 (30 min) | time window; 0 disables this dimension |
| `inject_cooldown_prompts` | `MEMINI_INJECT_COOLDOWN_PROMPTS` | 3 | prompt-count window; 0 disables this dimension |
| `inject_pretool_gate_ms` | `MEMINI_INJECT_PRETOOL_GATE_MS` | 90000 (90 s) | skip the per-file PreToolUse recall call entirely when that file's last call is younger; 0 restores always-call |

The suppression rule, shared by every surface: an injected memory stays suppressed while it is inside **either** window, and becomes eligible again once **both** have lapsed. On re-injection its clock and counter restart. Setting both windows to 0 reproduces the exact forever behavior of #134, and `inject_dedupe=false` still turns everything off (including the call gate, whose clock lives in the dedupe state).

The prompt counter is a real "user messages" unit: the UserPromptSubmit hook bumps and persists it on every prompt, above the short-prompt and command gates and above the `recall` setting gate, so a `MEMINI_RECALL=0` setup with pretool recall still enabled cannot freeze the counter and silently regress to forever-dedupe. On hosts that never fire a per-prompt hook the counter stays at zero and the prompt dimension goes inert (time window still governs) rather than suppressing forever.

The injected-state file moves to a v2 shape, `{v, n, ids: {id: {h, at, n}}}`. v1 files migrate on read and persist as v2 within one prompt. Writes merge against the file (counter takes the max, per-id the newer stamp wins) so parallel hook processes cannot rewind the counter. Old plugin versions read a v2 file as empty rather than crashing.

Recall activity events now carry `detail.excluded_count` when the client asked the server to skip already-injected ids, so the feed can distinguish dedupe pressure from organic serving.

## Per-surface behavior

| surface | dimensions | notes |
|---|---|---|
| Claude Code plugin | time + prompts | content-aware: a `memory_update` to an injected memory still re-injects immediately on the per-file surface; ids the model read itself via the memini MCP tools stay suppressed for the whole session (re-pushing those is pure noise) |
| Codex (same plugin) | time only in principle | Codex fires neither UserPromptSubmit nor the file-matcher PreToolUse hooks, so its dedupe surfaces are the briefing and MCP tool-read tracking; documented in the plugin README |
| hermes | time + prompts (prefetch count) | id-only, no content hashes, so no update bypass and no immortal entries; the exclude_ids 400 fallback now latches only when `exclude_ids` was actually in the failed request, closing a false-latch the windowing would otherwise have introduced |
| openclaw | time only | its injection hook runs per agent step, not per user message, so a prompt counter would count tool calls; `inject_cooldown_ms=0` there means forever-suppress, the safe reading for a per-step hook |
| pi | time + prompts | per-prompt hook, so both dimensions apply |

opencode and openwebui remain stateless, as in #134.

One layering note worth knowing before reading the pretool code: the per-file fingerprint from #134 (same file, same served set, skip the block) is deliberately unchanged, so a file being edited in a tight loop never gets an identical block re-pushed even after the windows lapse. Re-reminding flows through the prompt-recall surface and through recalls on other files; the fingerprint remains a per-file display dedupe.

## Testing

Every change was written test-first against the existing harnesses: the plugin suite grew from 156 to 181 (state migration, merge-on-write, a full predicate truth table, gate behavior including the no-network assertion, counter bumps with recall off, sentinel semantics), hermes from 66 to 75 (windowed exclude_ids, re-admission, the latch regression), openclaw 145 and pi 46 with clock-skew re-admission tests on both. The Go suite covers the new knobs through defaults, validation, merge, wire round-trip, and store conformance, plus the `excluded_count` event detail. All regen gates (`client-check`, `rest-check`, `gen-api-check`, `docs-check`) are clean at HEAD.

A live end-to-end pass against a real server (real hooks, real state files) confirmed: an Edit burst on one file produces exactly one recall event inside the gate; a memory injected on one prompt is excluded on the next and comes back after the windows lapse; both-zero knobs reproduce the forever behavior of #134.


eleboucher:
> _Originally by **eleboucher** on 2026-07-18 09:03 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/136#issuecomment-12958)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #137), which is the source of truth. Further review and merging happen there.

PR #56, opened 2026-07-19
Server-enforced floor on the served score, visible filtering in the activity feed, and pretool exclude_ids

perfectra1n:
## The problem

Two related issues showed up on a production deployment running v0.7.5 with the cross-encoder reranker and both existing gates configured.

First, the activity feed makes working dedupe look like spam. The feed logs every recall serve on the server, but the Claude Code plugin suppresses re-injection client side. A session that reads six files in two minutes produces six "served 3 memories" events full of repeats, while the actual conversation context received exactly one injection (verified against the session transcript). Nothing in the feed indicates that suppression happened.

Second, the score users see has no floor. The response `score` field, the activity feed, and the UI all carry the composite score (fused score max-normalized per pool, blended with quality). `MEMINI_RECALL_MIN_SCORE` gates the raw fused score before reranking, and `MEMINI_RERANK_MIN_SCORE` gates the cross-encoder scale, which never reaches the wire. Neither touches the composite scale, so with `limit: 3` the tail slots routinely fill with weakly related results. Replaying the activity log of that deployment: 27% of pretool serves scored below 0.5, concentrated at ranks 2 and 3 (rank 1: 15%, rank 2: 29%, rank 3: 40%).

There was also a scale bug hiding in the client knobs: `inject_pretool_min_score` and `inject_recall_min_score` were forwarded to the server as `min_score` (fused scale) while also being applied client side against the response score (composite scale). One knob value, two unrelated meanings, and each integration did it slightly differently.

## What this PR does

**1. `min_rank_score`: a server-enforced floor on the composite score** (`feat: add server-authoritative min_rank_score composite floor`)

New optional per-call search param, domain [0,1). It runs after the cross-encoder reorder and floors the final composite score, so it gates exactly the number the response and the feed display. Results below the floor are dropped from the response and never reinforced, but they are not silently discarded: they are logged to the activity feed as rows carrying a `filtered: "rank_floor"` marker with their true rank and score. Every logged row is numbered against the single pre-floor finalized list, so served rows show holes exactly where floored rows sit, and the UI renders floored rows dimmed with a badge. You can see precisely where the floor bit and what it cost.

Because the composite score is pool-relative (the top hit lands near the ceiling by construction), the floor trims trailing filler but can never empty a response on its own; detecting "no true answer at all" remains the cross-encoder gate's job. The floor is deliberately per-call rather than a server-wide env: hook-driven injections want it strict, while agent-driven MCP recalls often want the low-scoring tail (exploratory recall legitimately uses 0.2-0.4 hits). Linked-expansion results are exempt, since `expandLinked` assigns a synthetic half-of-minimum score that a floor would nullify. With `query_rewrite` the floor applies per variant, and the final RRF fusion is never re-floored (different scale).

**2. The Claude Code plugin knobs now mean one thing** (`fix(plugin): floor hook injection on server-side min_rank_score`)

The pretool and prompt hooks send the knob value as `min_rank_score` and no longer send `min_score` at all. The client-side composite filter is kept only as a fallback: the server rejects unknown fields (`DisallowUnknownFields`), so against an older server the hook retries once with the new fields stripped and only then filters client side. Values at or above 1 are clamped to client-only filtering instead of failing every search. A knob of 0 stays a no-op everywhere.

**3. Pretool sends `exclude_ids`, with a latch that preserves update resurfacing** (`feat(plugin): pretool latches exclude_ids after one unchanged re-serve`)

The prompt hook already excludes recently injected ids server side; pretool deliberately did not, so that a memory whose content changed could resurface via the hash check. That choice is what kept the feed full of repeats: the same top memories were served, logged, and then suppressed client side on every file read. The new design keeps both properties. The first re-serve of an injected id is still allowed through so the hash check can detect content updates. If that re-serve arrives with unchanged content, the id is latched (new additive `r` counter in the v2 session state) and rides `exclude_ids` until its cooldown windows lapse. Excluded ids never rank, never reinforce, and never appear as served rows; the event's `excluded_count` shows up as a chip in the feed. Old plugins simply drop the `r` field and keep today's behavior; no state version bump.

**4. Every bundled integration ports to the same semantics** (`fix(pi)`, `fix(opencode)`, `fix(hermes)`, `feat(openclaw)`)

pi, opencode (v1 and v2), and hermes all forwarded the same knob as fused-scale `min_score`; openclaw ignored the knob entirely. All of them now send `min_rank_score` under the same rules as the Claude plugin: sent only when the knob is in (0,1), clamped to client-only at or above 1, one 400 retry stripping the newer fields (`min_rank_score`, and `exclude_ids` where the integration sends it), and the client-side composite filter demoted to a fallback that runs only when the floor did not reach the server. opencode v2 previously had no retry path at all and gains one; openclaw gains the `inject_recall_min_score` knob end to end (schema, resolution, wire, fallback) and still never sends `min_score`. pi's `memory_recall` tool also gains `min_rank_score` to stay in parity with the server's MCP recall contract.

One deliberate non-port: the Claude plugin's first-re-serve-allowed latch exists so its content-hash check can resurface updated memories. The other integrations do not track content hashes, so allowing the first re-serve would buy nothing and only weaken their dedupe. They keep their existing exclude-all-in-window behavior, which already stops server-side repeat serves. openwebui (stateless filter, no floor valve) and codex (config only) are not affected.

## Compatibility

- Old server + new plugin or integration: one 400 retry without the new fields, client-side filtering takes over, no behavior regression.
- New server + old plugin: `min_rank_score` is optional; nothing changes until a client sends it.
- Existing state files: `r` is additive; the migration path and merge-on-write preserve it, with a round-trip test.

## CI note

The `plugins` job currently fails on main itself: the release process bumps `integrations/pi/plugin/package.json` but not its `package-lock.json`, and pi's "package metadata and npm lockfile root stay synchronized" test catches the mismatch (v0.7.5 and v0.7.6 both hit it). This PR syncs the lockfile via `npm install --package-lock-only` (two version fields, no dependency churn), so the job is green here. The release tooling probably wants the same sync step.

## Notes for review

- **Knob semantics change.** Anyone who tuned `inject_pretool_min_score` / `inject_recall_min_score` against the fused scale should re-tune against the composite scale (the score visible in the activity feed). Called out in the regenerated docs as well.
- **Sub-floor previously-injected memories never latch.** A memory that was injected once and later scores below the floor is dropped from the response before the hash check runs, so it cannot latch into `exclude_ids`. It recurs as a dimmed floored row instead. That is intentional (the floor decision stays visible), just calling it out so it does not read as a bug.
- **All-floored events carry the marker at event level.** When every hit is floored, the group head row is a floored row, so the operation-level detail includes `filtered`. Harmless (the UI reads only `source` and `excluded_count` there), noting it for completeness.
- **`exclusiveMaximum` stays in boolean form.** The pinned oapi-codegen/kin-openapi toolchain models `exclusiveMaximum` as a bool and fails generation on the numeric 3.1 form. The [0,1) bound is enforced server side in Go either way; the spec annotation is documentary.

## Testing

- 9 new Go service tests covering the floor: filtering and order preservation, zero/absent no-op, out-of-range rejection, placement after the cross-encoder (a recording fake reranker proves the pool is untouched and survivor order is kept, with holes), empty-verdict interaction, linked-expansion exemption, event rows carrying true rank/score plus the marker, and no reinforcement of floored hits. REST round-trip tests for the new field, including the 400 on invalid values.
- Claude Code plugin: 185 hook tests passing (9 new or rewritten), covering the wire bodies, the 400 retry with client-side fallback, the clamp path, the full latch lifecycle, the state-file `r` round trip through the disk merge, and the cross-surface path where a pretool-latched id rides the prompt hook's `exclude_ids`.
- Integrations, each with wire-body assertions that `min_rank_score` is sent and `min_score` is not, retry-strip tests, and server-enforced-path tests proving below-floor hits from an authoritative server are kept: pi 95 + 3 + 12, opencode 85, hermes 77, openclaw 77 + 70. All green.
- `mise run test`, `test-hooks`, `lint`, `rest-check`, `docs-check`, `gen-api-check`, and `client-check` all green.

We plan to run this deployment with both knobs at 0.5, keeping `MEMINI_RERANK_MIN_SCORE` at 0.2 for the absolute no-answer gate.


eleboucher:
> _Originally by **eleboucher** on 2026-07-19 10:24 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/140#issuecomment-13073)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #159), which is the source of truth. Further review and merging happen there.

PR #57, opened 2026-07-19
feat(server): measure HTTP response body size per route

perfectra1n:
Adds a `memini_http_response_bytes` histogram (chi route pattern + method, buckets 256B–256K) observed from the bytes the wrapped ResponseWriter already counts, plus a p50/p90 dashboard panel for the recall and briefing routes.

**Why:** the hook-driven injection pipeline ships several MB of `/v1/search` payload per day (measured on a production deployment: mean 6.8KB, p90 11.8KB per search response, ~1,300 searches/day), but only request counts and latency were measured. Response volume — the thing injection tuning changes — was invisible. This is the baseline metric for the injection-efficiency series.

Standalone; no behavior change. The series: #58 (directive slimming) → #59 (injection telemetry) → #60 (progressive disclosure) → #61 (server budgets) → #62 (enforcement core) → #63 (integration parity), plus #64 (floors, stacked on #56) and #65 (capture hygiene) — each independently mergeable in order.

PR #58, opened 2026-07-19
feat(plugin): slim the memory directive and emit it per fire source

perfectra1n:
Two cuts to SessionStart's fixed overhead:

**The directive shrinks ~2,070 → ~640 chars.** It was a full copy of the save policy whose canonical text already ships in the MCP server instructions every session (`mcp.go` serverInstructions — the sync comment in `_shared.mjs` already named it canonical). The hook copy is now explicitly the abridged trigger; one source of truth.

**Emission follows `payload.source`** instead of firing on every path:
- `startup` / `clear` / no source (Codex) → slim directive (fresh context)
- `resume` → **nothing** — Claude Code replays previously injected hook text on resume, so re-emitting was pure duplication
- `compact` → the compact-recovery nudge only — the MCP instructions persist in the system prompt across compaction

Net: ~72% less directive cost on startup, zero on resume/compact. Verified e2e (real server + hooks): startup output drops to ~1.4KB total, resume is byte-silent, compact re-briefs with the nudge alone.

Plugin-only, negative LOC. Stack base for the injection-efficiency series; #57 is independent.

### DIFF
--- a/plugin/scripts/_shared.mjs
+++ b/plugin/scripts/_shared.mjs
@@ -1257,49 +1257,25 @@ export function readToolCall(payload) {
 // keeps a legacy scraper (parseMemoryBlocks) as a back-compat fallback for
 // sessions still emitting inline <memory> blocks under the old instruction.
 //
-// The save-policy invariants below (what is durable, visibility, correction
-// hygiene) are canonical in internal/api/mcp/mcp.go serverInstructions; keep
-// the two phrasings in sync.
+// This is deliberately the ABRIDGED trigger, not the full save policy. The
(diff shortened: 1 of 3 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-07-19 10:25 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/143#issuecomment-13075)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #157), which is the source of truth. Further review and merging happen there.

PR #59, opened 2026-07-19
feat: injection telemetry — report what hooks actually inject, and show it

perfectra1n:
**Stacked on #58** (diff includes it until that merges).

The activity feed logs recall serves BEFORE the client's dedupe/cooldown gates run, so it over-reports what reached model context — a session reading many files shows "served 3 memories" rows for injections that never happened. Filtering was invisible; this closes the loop.

- **`POST /v1/activity/injected`**: after composing stdout, each injecting surface (briefing / prompt / pretool) sends one bounded beacon — injected ids, token/char estimates, suppressed counts by reason. `AbortSignal` 500ms, awaited before exit, errors swallowed, never touches stdout. Knob `inject_telemetry` (`MEMINI_INJECT_TELEMETRY`), default on, served via the handshake.
- **Server**: new `inject` event kind through the best-effort event funnel; `memini_injected_memories_total{surface,result}` + `memini_injected_tokens_total{surface}` counters; the activity regrouping joins reports onto nearby recall serves and annotates each served memory `injected: true/false` — **absent means unreported**, so old data and non-reporting integrations render unchanged.
- **UI**: recall rows read "served N → injected M", reported-suppressed memories get a `not injected` chip, inject events show surface + per-reason counts + token estimate. Grafana gains an Injection efficiency row.

Join caveat (documented in code): recall events carry no session id, so the join is namespace + 5-minute window + id membership — annotation appears when inject rows are in the fetched page.

Tests: 7 new hook tests (beacon shapes, hanging/500 endpoint never fails a hook), Go tests for event kind, join true/false/absent, handler validation, metrics.


eleboucher:
> _Originally by **eleboucher** on 2026-07-19 10:27 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/142#issuecomment-13138)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #161), which is the source of truth. Further review and merging happen there.

PR #60, opened 2026-07-19
feat: progressive disclosure — concise wire, [m:id] handles, prefix memory_get

perfectra1n:
**Stacked on #59** (diff includes #58/#59 until they merge).

Injection payloads switch from truncated-content blobs to an index the model can expand on demand (the claude-mem "index now, detail on demand" shape).

- **Server**: shared `internal/api/render` package projects concise content (summary if present, else a word/sentence-boundary cut — no more mid-sentence `…` artifacts) behind `response_format=concise` on `/v1/search` and `format=concise` on the briefing. Briefing gains `children=summary|full|none` — `summary` finally matches the isolation contract the docs claimed (titles/counts only); the admin UI keeps `full`.
- **`content_hash`** (16-hex sha256 of full content||summary) rides every memory in every format. Client dedupe identity prefers it, so a memory served full in the briefing and concise in a recall dedupes as the **same** memory — the regression that made a concise wire unsafe before.
- **`[m:<id8>]` handles** on every rendered bullet + a one-line teach header (only on blocks that actually truncated): `memory_get` — and `GET /v1/memories/{id}` — now resolve unambiguous id prefixes ≥8 hex chars (409 lists collisions; mutations stay verbatim-id-only, pinned by tests).
- Briefing Recent drops to strict index mode (120 code points + age + handle). Tail drops say what they dropped: `[+N more — memory_recall for detail]`.

Old servers: hooks fall back via the existing newest-field-first 400 retry chain; unknown query params are ignored.

Wire p90 for search responses drops roughly 4× on the production deployment this was measured against.


eleboucher:
> _Originally by **eleboucher** on 2026-07-19 10:25 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/141#issuecomment-13074)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #158), which is the source of truth. Further review and merging happen there.

PR #61, opened 2026-07-19
feat: server-enforced token budgets on recall and briefing

perfectra1n:
**Stacked on #60.**

The `inject_*_max_tok` knobs stop being silent client-side trims and become request budgets the server honors — filtering happens where it's visible:

- `/v1/search` gains `max_tokens`: fill in final rank order, first result always fits (a non-empty recall never becomes empty by budget), response reports `omitted`. Recall activity events record `budget_omitted`.
- Briefing gains `?max_tokens`: fill pinned → facts → procedures → recent, whole items only (fill order IS priority order — recent starves first, pinned never), response reports `omitted`.
- Estimation mirrors the client's `approxTokens` recipe over the content the response actually ships (concise text in concise mode) + a 10-token per-item skeleton overhead.
- Hooks pass the knobs on the wire and render the server's count in the `[+N more — memory_recall for detail]` footer; client-side `fitByTokens` stays as the old-server fallback (400 strips `max_tokens` first in the retry chain).
- Defaults come off unbounded: briefing 600 / recall 250 / pretool 200 — sized so the median injection is untouched and the multi-KB tail is what gets clipped.


eleboucher:
> _Originally by **eleboucher** on 2026-07-19 10:26 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/150#issuecomment-13081)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #151), which is the source of truth. Further review and merging happen there.

PR #62, opened 2026-07-19
refactor: single injection-enforcement core in @memini/client, pinned by golden vectors

perfectra1n:
**Stacked on #61.** Behavior-preserving — zero test-expectation changes across the plugin (202) and client (105 pre-existing) suites.

The injection-limiting primitives (suppression windows, seen-set merge, token budgeting, render formats, identity/fingerprints, beacon payload) lived as hand-written .mjs in the plugin and were copy-pasted, drifting, per integration. They now live once as pure functions in `packages/memini-client/src/enforce/` (no fs/env/clock — state and `now` are parameters), bundled through the existing `_client.gen.mjs` path; `_shared.mjs` keeps only the Node adapters and re-exports the same names.

`vectors/enforcement.json` pins 54 golden cases generated once from the original implementations — the cross-language contract future ports are tested against (the hermes Python port uses it in the next PR).

pi switches to the core; two real drifts surfaced and were adopted (its identity lacked the content_hash preference, its fitByTokens lacked the partial-fit trim), while pi-specific render wordings its tests pin stay local.


eleboucher:
> _Originally by **eleboucher** on 2026-07-19 10:26 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/149#issuecomment-13080)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #152), which is the source of truth. Further review and merging happen there.

PR #63, opened 2026-07-19
feat(integrations): windowed injection-cooldown parity everywhere

perfectra1n:
**Stacked on #62** (last of the stack).

The enforce core's semantics now hold across every integration:

- **openwebui** had no dedupe at all — it re-served identical memories every turn. It gains the standard per-chat windowed cooldown (handshake-served knobs, 30min/3-prompt defaults), `exclude_ids` wire exclusion with the retry-once-and-latch old-server fallback, and the content-identity bypass.
- **opencode** suppressed forever within a session — the attention-decay failure the windowed design exists to avoid. Both shipped entry points move to windowed entries; **behavior change flagged in its README** (0/0 restores forever).
- **openclaw**'s prompt-count window was inert. It now counts completed agent turns (`agent_end`) — the once-per-turn signal that host can honestly maintain; hosts without it degrade to time-only, never forever.
- **hermes** tracked ids without content identity, so an in-place-updated memory stayed withheld for the full window. It adopts the content_hash identity + mismatch bypass, and gains a golden-vector runner replaying `vectors/enforcement.json` — the Python drift gate.

Suite counts: opencode 84→115, openwebui 37→48, hermes 74→84, openclaw 75+70→77+72; plugin/client/pi untouched.


eleboucher:
> _Originally by **eleboucher** on 2026-07-19 10:25 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/148#issuecomment-13079)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #153), which is the source of truth. Further review and merging happen there.

PR #64, opened 2026-07-19
feat: raise injection floors to 0.5 (server-enforced) and stop recalling on Glob/Grep

perfectra1n:
**Stacked on #56 (rank-floor)** — the diff shows its commits until #56 merges, and this must not ship in a plugin release before #56's client does: a pre-#56 bundle interprets a served 0.5 on the fused scale and over-filters badly. Escape hatch either way: `MEMINI_INJECT_*_MIN_SCORE=0` locally (env beats server).

Live measurement showed the floors filtering almost nothing: 95.8% of recalls returned 2–5 hits and only 1.7% came back empty, so 0.1–0.46-relevance matches rode every hook injection.

- `inject_recall_min_score` / `inject_pretool_min_score` default 0 → **0.5**, enforced **server-side** as #56's composite post-rerank `min_rank_score` gate. What the floor drops is visible, not silent: recall events record `floored=N`, floored rows are marked, and a `memini_recall_floored_total` counter tracks it.
- `inject_pretool_tools` drops **Glob/Grep**: their recall queries embed the search pattern ("Grep on <pattern>") — near-zero semantic signal, and each ungated call costs a server embed+rerank (several hundred/day at observed volume). The knob restores them; the hook matcher still carries both so the knob alone re-enables.
- `MEMINI_RECALL=0` now gates the PreToolUse recall too (it previously kept firing with the master switch off).


eleboucher:
> _Originally by **eleboucher** on 2026-07-19 10:28 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/147#issuecomment-13139)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #160), which is the source of truth. Further review and merging happen there.

PR #65, opened 2026-07-19
feat(service): strip harness boilerplate from turn captures at write time

perfectra1n:
Standalone.

Auto-captured conversation turns (metadata `format=turn` — the predicate every integration stamps) arrive polluted with harness noise the model echoed or the hooks injected: the plugin's own `<memini-*>` wrapper blocks, `<system-reminder>` blocks, and output-style `★ Insight` banners. Stored as-is, recall re-serves that boilerplate as memory content — observed live.

Strips whole blocks server-side in `Remember`, before the episodic value gate, so a boilerplate-only capture shrinks under the min-chars gate and is dropped instead of stored; a capture emptied entirely drops outright. Non-capture writes are never touched. Keying on `format=turn` rather than the Claude-only `source=turn_capture` is what makes it land for every integration.

### DIFF
--- a/internal/service/capture_hygiene.go
+++ b/internal/service/capture_hygiene.go
@@ -1,0 +1,122 @@
+package service
+
+import (
+	"regexp"
+	"strings"
+
+	"github.com/eleboucher/memini/internal/memory"
+)
+
+// Turn captures — conversation turns auto-saved by client integrations — are
+// identified by isTurnCapture (metadata format="turn", see service.go): every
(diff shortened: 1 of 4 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-07-19 10:25 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/146#issuecomment-13078)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #154), which is the source of truth. Further review and merging happen there.

PR #67, opened 2026-07-19
Server rank floor, 0.5 defaults, shared enforcement core, and integration cooldown parity

perfectra1n:
This combines the four PRs from the injection series that were still open after the token-budget stack landed (your 151/154/156): #56 (server rank floor), #62 (shared enforcement core), #63 (cooldown parity for the integrations), and #64 (0.5 floors, no more Glob/Grep recall). They were developed as two parallel stacks that both rewrite the client enforcement code, so rebasing each one separately would have handed you four rounds of overlapping conflicts. This branch reconciles them once, on top of current main, with everything re-tested against the merged result.

## Why these still matter

The merged budget stack caps how much an injection can ship. This half of the series decides what deserves to ship at all:

* The floor moves to the server. `min_rank_score` floors the final composite score (the one the activity feed shows), so one knob value means the same thing in every integration, and floored hits stay visible in the feed (dimmed, marked as filtered) instead of silently vanishing client-side.
* The floor default goes to 0.5 and pretool stops recalling on Glob/Grep. On the live deployment, pattern-derived queries ("Grep on `<pattern>`") were the bulk of near-zero-signal serves, and each one costs a server embed plus rerank.
* The enforcement primitives (identity hashing, the cooldown predicate and its state algebra, token budgeting, rendering, telemetry) move into `packages/memini-client/src/enforce/`, pinned by golden vectors, so pi/openclaw import them directly and the hooks consume the committed bundle. One implementation instead of five drifting copies.
* The integrations get the same windowed cooldown the Claude Code hooks already have: a memory suppressed while in its time or prompt window, re-admitted when both lapse, content changes always resurfacing.

## Where combining them was real work (worth a closer look in review)

* The pretool exclude_ids latch (#56) was written against the old `_shared.mjs` function bodies, which #62 moved into `enforce/seen.ts`. The latch is implemented in the new home: `recordInjected` stamps `r: 0` and resets an earned latch, `pretoolExcludeIds` reuses the exact `cooldownIds` window predicate, and `normInjectedEntry` preserves `r` without backfilling it on read. That last bit is deliberate: the golden vectors were cut from the pre-latch implementation, and not backfilling keeps every vector byte-identical, so `vectors/enforcement.json` is untouched by this branch.
* Recall pipeline order is now finalize, floor, linked expansion, budget. The floor runs on the final score before expansion (linked hits carry a synthetic score a post-expansion floor would nullify); the budget trims last so it sees exactly the list the caller would receive. The activity event carries `floored`, `budget_omitted`, and `excluded_count` side by side, and the feed's recall headline composes both features ("served 2 -> injected 1", "served 0 / 3 floored").
* The old-server 400 fallback keeps the strip-one-field-per-retry chain from the budget stack, with `min_rank_score` joining at the front: min_rank_score, then max_tokens, then exclude_ids, then response_format. Since the 0.5 floor now rides every search, a server that predates it costs one extra retry on the first shape; every intermediate shape is valid, and a stripped floor falls back to the client-side filter.
* The opencode and openclaw floor ports predate the parity refactor, so they were re-seated onto the parity versions: v1 keeps its cooldown-derived exclude_ids, v2 and openclaw keep the windowed seen state, and openclaw wires the floor knob alongside both cooldown dimensions.
* Defaults merged rather than fought over: the landed token budgets (600 briefing / 250 recall / 200 pretool) stay, the min-score floors go to 0.5, Glob/Grep leave the default pretool allowlist (listing them restores the old behavior).

## Testing

* `go test ./...` green.
* Client core 180/0, enforcement vectors untouched and passing.
* Claude hooks 215/0, opencode 118/0, openclaw 151/0, hermes 87 OK, openwebui 48 OK.
* pi: 94 pass, 1 fail. The failing test ("precompact and shutdown checkpoints are bounded, gated, ...") fails identically on current main, so it is pre-existing and not from this branch. Happy to chase it separately.
* All generated artifacts (api.gen.go, `_client.gen.mjs`, UI catalog, docs reference) regenerated, zero drift.

Supersedes #56, #62, #63, and #64. The commits keep the original series structure (core, parity, floor stack, defaults) in case you want to review or land them piecewise anyway.


eleboucher:
> _Originally by **eleboucher** on 2026-07-19 16:36 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/162#issuecomment-13221)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #163), which is the source of truth. Further review and merging happen there.

PR #68, opened 2026-07-19
fix(activity): hydrate inject events with the serve's memory snapshot

perfectra1n:
## The problem

`inject` events render as blank rows in the activity feed, and clicking one 404s. Every injected-memory row on my instance was affected (104 of 104 came back with an empty `namespace`, `tier` and `summary`). Because `recall` and `inject` events interleave in the feed and recall rows render fine, it reads as an intermittent glitch rather than a total failure of one event kind.

A served row looks like this:

```json
{"id": "...", "namespace": "acme", "rank": 1, "score": 0.61, "summary": "...", "tier": "semantic"}
```

An injected row looks like this:

```json
{"id": "...", "namespace": "", "rank": 1, "summary": "", "tier": ""}
```

## Root cause

Event rows carry a denormalized snapshot of the memory (`memory_ns`, `memory_tier`, `memory_summary`), stamped at write time by `applyMemory`. `logRecallEvent` and `logBriefingEvent` both call it. `RecordInjected` never did, setting only `MemoryID` and `Rank`, so inject rows persisted three empty strings.

That single omission causes three separate symptoms:

1. The UI draws an empty line (and an unstyled `TierBadge` dot).
2. Inject events never match the `?tier=` or `?q=` activity filters, since both subqueries key off `memory_tier` and `memory_summary`. This one is silent and I do not think it has been reported.
3. Click-through 404s. The UI passes the row's empty namespace to `api.get`, and `ns ?? namespace.value` does not fall back on `""`, so the request goes out with no `X-Memini-Namespace` header at all.

## Why it is not a one-line fix

The obvious patch is to stamp the request namespace onto the row. That writes a plausible lie: recall serves across the ancestor read-set cascade, so a memory served to `acme/dev` may well live in `acme`. The click would still 404, now with wrong data on the row.

The beacon cannot supply it either. `injected_ids` is a list of bare id strings, and any fix has to work for already-deployed plugin versions.

Resolving ids against the memories table would work (ids are globally unique) but opens an information leak. Injected ids are taken on faith and never authorized, so a buggy or hostile client could beacon arbitrary ids and read back their summaries out of the activity feed.

## The fix

Borrow the snapshot from the serve that produced the injection, via a new `EventLogStore.ServedSnapshots`. That source is both the accurate one and the safe one:

- **Accurate**, because the serve row is the only place the memory's own namespace was ever recorded. The inject row ends up reading identically to the recall row above it in the feed.
- **Safe**, because the lookup is scoped to serves already logged against the caller's namespace. A report can only learn what it was already shown. An id no serve covered keeps its bare row, preserving the existing "taken on faith" contract.

Hydration runs inside the existing background log goroutine rather than inline, so it stays off the beacon's latency path and gives the preceding serve's own append a head start on landing.

Implemented for both backends over the `idx_memory_events_memory` index that already exists. No migration and no schema change.

`Score` is deliberately not copied. The composite score belongs to the recall that computed it against its query, and duplicating it would have two events claim the same number.

## UI

Server-side hydration fixes new events, but rows written before this change stay bare until they age out of the retention window, so the frontend degrades rather than blanking:

- `TierBadge` renders nothing for an empty tier. Fixed in the shared component, since `MemoryCard` and `MemoryDrawer` have the same exposure.
- An empty summary falls back to the id prefix, which memini can resolve, instead of an empty line.
- `api.ts` uses `||` rather than `??` so an empty namespace falls back to the active one instead of dropping the header.
- The silent `catch` around opening a memory now surfaces a message. Swallowing it is what made a 404 look like a dead button and hid this for as long as it hid.

## Tests

`TestRecordInjectedWritesEvent` existed and asserted only that the ids survived, never that a row was renderable, which is how this shipped. Added:

- `TestRecordInjectedHydratesServedSnapshot`: the inject row carries the same snapshot as the serve it reports on, and no score.
- `TestRecordInjectedHydratesAncestorNamespace`: a memory owned by a parent namespace and served to a child reports the parent. This is the assertion that distinguishes this fix from stamping the request namespace, and it is exactly the case that 404s today.
- `TestRecordInjectedLeavesUnservedIDsBare`: an id served only to another namespace, and one written but never served, both stay bare.
- Store conformance coverage for `ServedSnapshots` (newest-wins, snapshot-less rows are not donors, namespace scoping, kind filtering, the `since` bound, empty input). Runs against sqlitevec and postgres automatically.
- `TestReportInjected` now asserts the wire shape the dashboard actually reads.

I confirmed the new service tests fail without the fix, with the reported symptom (`ns:"" tier:"" summary:""`), rather than passing vacuously.

## Verification

- `go test ./...` green.
- Store conformance green on sqlitevec and on a real vchord-postgres container (`-tags integration`).
- `pnpm typecheck` and `pnpm build` green.
- End to end against a local server: seeded a memory in a parent namespace and one in a child, recalled from the child so the cascade served both, then POSTed a bare-id beacon exactly as the plugin does. Both rows came back fully populated, the ancestor-owned one correctly reporting the parent namespace, and fetching each memory with the namespace its row reports returns 200 for both. `?kind=inject&tier=semantic` and `?kind=inject&q=...` now match. Beaconing an id from an unrelated namespace still records the row and still leaks nothing.

## Note for the changelog

Hydrated inject events now match `?tier=` and `?q=`, so a tier-filtered or text-filtered feed will start returning inject events it previously omitted. That is the fix working, but it could read as a regression to someone who does not know the filters were broken for that kind.

### DIFF
--- a/internal/api/rest/activity_test.go
+++ b/internal/api/rest/activity_test.go
@@ -26,13 +26,14 @@ type activityEvent struct {
 	Query     string         `json:"query"`
 	Detail    map[string]any `json:"detail"`
 	Memories  []struct {
-		ID       string   `json:"id"`
-		Summary  string   `json:"summary"`
-		Tier     string   `json:"tier"`
-		Rank     int      `json:"rank"`
-		Score    *float64 `json:"score"`
-		Section  string   `json:"section"`
-		Injected *bool    `json:"injected"`
+		ID        string   `json:"id"`
+		Namespace string   `json:"namespace"`
+		Summary   string   `json:"summary"`
+		Tier      string   `json:"tier"`
+		Rank      int      `json:"rank"`
+		Score     *float64 `json:"score"`
+		Section   string   `json:"section"`
+		Injected  *bool    `json:"injected"`
 	} `json:"memories"`
 }
 
@@ -251,6 +252,20 @@ func TestReportInjected(t *testing.T) {
 		}
 		if len(ev.Memories) != 2 {
 			t.Fatalf("inject event carries %d memory refs, want 2 (unknown ids kept)", len(ev.Memories))
+		}
+		// The wire contract the dashboard renders from. A served id must come
+		// back with the snapshot of the serve that produced it — an empty one
+		// draws a blank row and leaves the UI no namespace to open it with —
+		// while an id nothing served stays bare.
+		byID := map[string]struct{ ns, tier, summary string }{}
+		for _, m := range ev.Memories {
+			byID[m.ID] = struct{ ns, tier, summary string }{m.Namespace, m.Tier, m.Summary}
+		}
+		if got := byID[injectedID]; got.ns != "acme" || got.tier != string(memory.TierSemantic) || got.summary == "" {
+			t.Errorf("injected ref = %+v, want a populated acme/semantic snapshot", got)
+		}
+		if got := byID["unknown-id-is-fine"]; got.ns != "" || got.tier != "" || got.summary != "" {
+			t.Errorf("unserved ref = %+v, want it left bare", got)
 		}
 	})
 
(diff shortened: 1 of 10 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-07-20 06:15 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/164#issuecomment-13307)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #173), which is the source of truth. Further review and merging happen there.

PR #73, opened 2026-07-27
feat(auth): add read-only API keys

perfectra1n:
Adds `read_only`, a second per-key authorization bit next to `admin`. A read-only credential reads everything it could read before, and the server refuses every mutating request it makes, on both the REST and MCP surfaces.

## The problem

I want to give a CI job's LLM access to memini. It should be able to recall project context, but it must never write, update, or delete a memory. Today there's no way to express that: every key that can authenticate can write, and `admin` only gates key management. A CI agent runs unattended, usually on a branch nobody has reviewed yet, and a bad write from one is invisible until it poisons somebody's recall weeks later.

## How it's enforced, and why this way

Both HTTP surfaces use the same shape: an explicit allowlist of what a read-only key may do, with everything else refused.

On REST that's a single middleware in `Mount`, registered right after `authMiddleware`: safe methods, plus three read-shaped POSTs (`/v1/search`, `/v1/answer`, and `/v1/handshake`, which is documented side-effect-free and is the call every client makes before anything else). On MCP it's a receiving middleware gating `tools/call` against an allowlist of read tools. MCP mounts outside `rest.Mount`'s group, so without that second gate a read-only key could just write over MCP instead.

The obvious alternative was to mirror the existing `requireAdmin` pattern and put a `requireWrite` call at the top of each mutating handler. I didn't, because that's about 20 call sites that fail *open*: add a write endpoint later, forget the line, and read-only keys silently get write access with no test to catch it. With an allowlist the default is refusal, so forgetting over-restricts (a loud 403 on something that should have been readable) instead of quietly granting writes.

Two tests keep the allowlists honest rather than relying on anyone remembering:

- `TestIsReadRequestClassifiesEverySpecEndpoint` parses `api/openapi.yaml` and asserts every `/v1` operation is classified, so adding an endpoint without deciding read-or-write fails CI.
- `TestMCPReadToolAllowlistMatchesAnnotations` cross-checks the tool allowlist against each tool's own `ReadOnlyHint` annotation. Those annotations were already authoritative, since `cmd/gendocs` builds the tool reference labels from them.

`read_only` is deliberately independent of `admin` rather than both collapsing into one `role` enum. `admin` plus `read_only` is a useful combination (an auditor who can list keys and read server defaults but change neither), and a role enum would have been a breaking wire, YAML, and CLI change for no gain.

## Trade-offs I made on purpose

These are choices, not oversights, so please push back on any of them:

- **The self-guard is stricter than admin's.** A named key can't make *itself* read-only; that returns 409. Once the flag is set, the gate refuses `/v1/keys` too, so the key could never lift it again. Clearing `read_only` on itself is allowed, since that direction is a restoration and is only reachable while the key can still write.
- **MCP refuses with an error tool *result*, not a protocol error.** A protocol error reads to an agent as "the call broke", which invites a retry. An error result is ordinary output the model sees and adapts to, and it says "Do not retry" in as many words. That matters for an unattended agent whose hooks would otherwise try a write every turn.
- **Write tools stay listed** for a read-only MCP session instead of being hidden from `tools/list`. Hiding them is a reasonable design too, so there's a test pinning the current behavior; switching would be a conscious change to what the agent sees, not a silent one.
- **Reads still reinforce.** Serving a read still bumps access counters and still appends to the activity log. That's internal relevance bookkeeping, and `read_only` bounds what the caller can change, not whether serving a read leaves a trace. A read-only agent's recalls still improve ranking, which I think is what you want.
- **Dry runs count as writes.** `POST /v1/namespaces/move` and `/split` post to a mutating endpoint even with `dry_run: true`, so a read-only key can't preview a move. The UI disables Preview alongside Apply rather than moving the 403 one click later.
- **Clients skip writes instead of collecting 403s.** `identity.read_only` rides the handshake the client already performs, so the plugin stops attempting captures rather than eating a 403 and logging it every single turn. An older server omits the field, and clients read absent as writable, so this can't make a client silently stop saving against an older server.

## One thing worth being explicit about

`read_only` bounds what a key can change, not what it can see. A read-only key can still send any `X-Memini-Namespace` and read every namespace on the server. I put a warning on that in `docs/api-keys.md` and again in the access-control guide, because someone reading "read-only" and inferring "contained" would be worse off than with no doc at all. Restricting what a credential can read is still a deployment topology decision.

That also meant rewriting the opening section of `docs/api-keys.md`. It used to say "there is exactly one authorization bit a key carries", which this change makes false, and the section was titled "Keys are identity, not authorization". It's now "not isolation", which is the accurate framing: a key does carry authorization, it just doesn't scope namespaces. The two places that echoed the old wording and every anchor pointing at the old heading are updated too.

## What changed

| Surface | Change |
| --- | --- |
| Storage | `store.APIKey.ReadOnly` on both drivers, `read_only` column defaulting to false so every existing key stays read-write |
| Auth | `apiauth.Principal.ReadOnly` from table and file keys. The env key and dev mode still resolve to a nil principal and stay fully privileged |
| REST | Fail-closed middleware, plus `read_only` on `ApiKey`, the create/update requests, and `CallerIdentity` |
| MCP | Call-time refusal via the read-tool allowlist |
| CLI | `memini key add --read-only`, a `READ ONLY` column in `key ls`, preserved across rotation |
| File keys | `read_only: true`, which is how a CI credential actually gets provisioned (a mounted Secret) |
| UI | Persistent topbar chip, per-key chip with a lock/unlock toggle, a create-form checkbox, and 15 write controls disabled with one shared explanation |
| Docs | A "The read-only attribute" reference section, the CI agent walkthrough, an fsck token caveat, and a chart values example |

## Verification

Every CI gate passes locally: `go test -race ./...`, `go build -tags=integration ./...`, golangci-lint with 0 issues, 219 hook tests, 180 client core tests, the UI build, and all five generated-artifact drift gates. Each commit is independently green.

I also checked it against a running server rather than only in tests. With a `--read-only` key, `POST /v1/memories` returns 403, `POST /v1/search` returns 200, `POST /v1/handshake` returns 200, and `DELETE /v1/pins` returns 403. The same holds for a key provisioned through a declarative `MEMINI_API_KEYS_FILE` entry, which is the path CI will really use, while a read-write file key and the break-glass env key on that same server are unaffected and still return 201. `GET /v1/self` reports `{"admin": false, "authenticated": true, "key_name": "ci-agent", "read_only": true}`.

For the UI I signed in with an `admin` plus `read_only` key: the chips render, every write control is disabled with the shared hint, and reads still work. I measured the new chip's contrast in the browser at 5.74:1 on the dark theme and 4.83:1 on the light one, both above the 4.5:1 floor for 11px text.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -723,7 +723,7 @@ paths:
     get:
       operationId: listApiKeys
       tags: [memory]
-      summary: List API keys (name/home/default namespace/created/disabled/admin/source — never a secret or hash)
+      summary: List API keys (name/home/default namespace/created/disabled/admin/read_only/source — never a secret or hash)
       description: >
         Admin-gated: allowed for the admin env key (MEMINI_API_KEY), a named
         key with admin=true, or dev/bootstrap mode (auth disabled entirely —
@@ -777,7 +777,7 @@ paths:
     patch:
       operationId: updateApiKey
       tags: [memory]
-      summary: Update an API key's home namespace, default namespace, disabled state, admin capability, and/or per-key settings
+      summary: Update an API key's home namespace, default namespace, disabled state, admin capability, read-only capability, and/or per-key settings
       description: >
         Admin-gated, see listApiKeys. Preserve-unspecified semantics matching
         `memini key add`'s rotation contract: an omitted field leaves the
@@ -2100,7 +2100,7 @@ components:
         delete all reject a "file" key with 409.
     ApiKey:
       type: object
-      required: [name, disabled, admin, source]
+      required: [name, disabled, admin, read_only, source]
       properties:
         name: { type: string }
         home:
@@ -2127,6 +2127,16 @@ components:
             exactly like the admin env key (MEMINI_API_KEY). A self-guard
             still applies — an admin key cannot demote, disable, or delete
             itself over this API.
+        read_only:
+          type: boolean
+          description: >
+            Whether this key is a read-only credential: it may issue reads
+            (GET/HEAD, plus the read-shaped POSTs /v1/search, /v1/answer and
+            /v1/handshake) and every mutating request is refused with 403,
+            across both the REST and MCP surfaces. Independent of admin — an
+            admin key that is also read_only may enumerate keys but not change
+            them. Note this bounds what the key may CHANGE, not what it may
+            SEE: a read-only key can still name any namespace and read it.
         source: { $ref: "#/components/schemas/ApiKeySource" }
         settings:
           allOf:
@@ -2169,6 +2179,12 @@ components:
           description: >
             Create the key as an admin credential (see ApiKey.admin). Defaults
             to false — a plain key that cannot reach the admin-gated surfaces.
+        read_only:
+          type: boolean
+          default: false
+          description: >
+            Create the key as a read-only credential (see ApiKey.read_only).
+            Defaults to false — a key that may write.
     UpdateApiKeyRequest:
       type: object
       properties:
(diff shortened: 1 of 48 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-07-27 20:36 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/185#issuecomment-13901)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #186), which is the source of truth. Further review and merging happen there.

PR #75, opened 2026-08-01
feat(llm): configurable completion budget and raw extra-body passthrough

KorenKrita:
## Motivation

The chat clients send `temperature: 0` + `max_tokens: 4096` (the built-in `defaultMaxTokens`). That is exactly right for the recommended non-reasoning models — but on OpenAI-compatible gateways many popular models now reason by default and **spend the completion budget on hidden thinking**.

Measured with `deepseek-v4-flash` on the distill prompt (real episodes, memini-identical request):

| | latency | completion tokens | of which reasoning | finish_reason |
|---|---|---|---|---|
| default request | 32.7s | 4,097 | 3,966 | `length` — JSON truncated, distillation lost |
| same + `{"thinking":{"type":"disabled"}}` | 5.4s | ~500 | 0 | `stop`, identical output quality |

The failure is silent in practice: distill-on-write only warns, so a reasoning model quietly loses most write-time distillations while everything looks healthy.

## What this adds

Two additive env knobs, both defaulting to current behavior:

**`MEMINI_LLM_MAX_TOKENS`** (int, default 0 = keep built-in 4096) — the dialect-free fix: give reasoning models enough headroom that the answer survives the thinking. Validated `>= 0` at config load.

**`MEMINI_LLM_EXTRA_BODY`** (JSON object, default unset) — merged into every chat request at the top level, for provider dialect knobs memini deliberately does **not** model:

```sh
MEMINI_LLM_EXTRA_BODY={"thinking":{"type":"disabled"}}   # DeepSeek-style
MEMINI_LLM_EXTRA_BODY={"enable_thinking":false}            # Qwen-style
MEMINI_LLM_EXTRA_BODY={"reasoning_effort":"low"}          # effort-style gateways
```

The reasoning-toggle dialect landscape is too fragmented to enumerate (DeepSeek `thinking`, Qwen `enable_thinking` / `chat_template_kwargs`, per-vendor `reasoning_effort` level sets…), so this stays a raw passthrough: one setting, one meaning, and the dialect choice remains the operator's. Fields memini sets itself (`model`, `messages`, `temperature`, `max_tokens`) always win — implemented with the SDKs' own `option.WithJSONSet`, applied to both the OpenAI and Anthropic clients. Invalid JSON fails config loading.

## Notes

- `go test ./internal/...` passes; new test `TestOpenAIExtraBodyMergedIntoRequest` asserts the merge reaches the wire and client-set fields survive
- `gen-docs` regenerated (`docs/reference/configuration.md`), group table updated
- No behavior change when neither variable is set

### DIFF
--- a/cmd/gendocs/main.go
+++ b/cmd/gendocs/main.go
@@ -110,7 +110,7 @@ var groups = []group{
 			"handling all work. Configuring one adds background consolidation, `POST /v1/answer`, the `memory_answer` " +
 			"MCP tool, and `MEMINI_RERANK=llm`.\n\nNote that `memory_answer` is only registered as an MCP tool when " +
 			"an LLM is configured. Without one, agents will not see the tool at all.",
-		Vars: []string{"MEMINI_LLM_BASE_URL", "MEMINI_LLM_API_KEY", "MEMINI_LLM_MODEL", "MEMINI_LLM_API"},
+		Vars: []string{"MEMINI_LLM_BASE_URL", "MEMINI_LLM_API_KEY", "MEMINI_LLM_MODEL", "MEMINI_LLM_API", "MEMINI_LLM_MAX_TOKENS", "MEMINI_LLM_EXTRA_BODY"},
 	},
 	{
 		Key:   "rerank",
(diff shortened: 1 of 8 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-08-02 18:52 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/189#issuecomment-14595)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #192), which is the source of truth. Further review and merging happen there.

PR #76, opened 2026-08-02
Make a degraded write non-destructive, and stop secondary failures from failing accepted writes

perfectra1n:
## The problem

A `memory_remember` that cannot reach the embedder is already stored vectorless and re-embedded later. What it also does, silently, is skip every write-time job that needs a query vector (the dedup gate, corroboration routing, contradiction routing, and consolidation), and `BackfillEmbeddings`' own doc comment says those are *"deliberately NOT re-run"* on repair.

So an embedder blip does not delay that enrichment, it **loses** it, leaving duplicate and contradictory memories that nothing ever reconciles.

Tracing that turned up two more problems worth fixing on their own:

- **The repair loop had head-of-line blocking.** `BackfillEmbeddings` aborts the entire tick when the *first* row fails, and there is no attempt counter and no next-run time. One permanently-unembeddable memory that happens to sort first wedges every other repair **forever**, silently, while logging a warning that blames the embedder.
- **An accepted write could be lost to a client disconnect.** `store.Upsert` ran on the raw request context. Every driver's Upsert is a transaction, and `database/sql` and pgx both roll one back on cancellation, so a client hanging up between "we decided to store this" and the commit lost the memory and returned a 500 for it.

## What this does

**Repair state moves onto the memory row** (`embed_state`, `embed_attempts`, `embed_next_run_at`, `embed_last_error`, plus a partial index) rather than into a job table.

That choice is load-bearing rather than incidental. `metadata.pending_embed` was already written inside `Upsert`'s transaction, so the memory row *is* the outbox record; a separate job table would have reintroduced exactly the dual write the pattern exists to prevent, and needed a reconciler to paper over the resulting lost-enqueue window. On the row, the state and the write commit together and that window does not exist. It also means every transition is a targeted `UPDATE` that never touches `embedding`, which structurally removes the `Get`-then-`Upsert` vector-loss trap documented on `GetEmbedding`: the repair path has no way to express "drop the vector". And it replaces a per-namespace `json_each` / `jsonb` full scan with a real index.

**A degraded write now loses nothing but time:**

```
pending --(embed)--> enrich --(replay the write-time jobs)--> healthy
   |                    |
   +---(N failures)-----+--> failed --(sweeper, after a rest)--> pending
```

The enrich stage replays what the write missed. Corroboration and contradiction routing are the same detached functions the write path calls. The only reason they were skipped is that the memory had no vector. Dedup is replayed too, with the one unavoidable difference that the row already exists: where a healthy write coalesces *before* storing, a repair supersedes instead of deleting, so the ID the caller was given still resolves to the surviving fact. The merge hint cannot be recovered (it was a field on a response already sent); the activity log records the merge in its place.

Backoff is exponential with full jitter, deliberately gentler than the `attempt^4` schedules general-purpose job queues use, which reach days by the twentieth attempt, which for an embedding backlog means a memory stays unsearchable for a week after a transient blip. Exhausting the ceiling parks the row with its error rather than dropping it, and the sweeper re-arms it after a rest, so a multi-hour outage costs a couple of probes an hour instead of continuous hammering and recovery needs no human.

**Nothing else can fail an accepted user action.** `store.DurableCtx` detaches an already-decided write from the request context (applied to writes only; reads and bulk passes must still die with their request). `CreateApiKey`/`RotateApiKey` stop reading the key back just to build the response, which used to 500 a key that *had* been created and lose its one-time secret forever. `Recall` and `Briefing` stop failing all-or-nothing across the read set: `Recall` used `errgroup.WithContext`, so the first leg to fail cancelled every sibling and took down a recall the primary namespace had already answered.

The rule throughout is that the **primary** namespace failing is fatal while a secondary leg failing is a degrade. A recall returning nothing from the namespace you asked about is a failed recall, not a degraded one, and returning empty would produce exactly the confident negative agents are warned about. For the same reason in reverse, the handshake's **pin** lookup deliberately stays fatal: a pin exists precisely because derivation gets the namespace wrong, and the client treats a successful handshake as authoritative, so degrading it would silently misfile every write of the session.

**A stuck repair is now visible.** Until now a repair that had given up rendered identically to one that would finish in two seconds, which is how permanent degradation hides. `/healthz?verbose=1` grows a `repair` block, `/v1/stats` grows `embed_stuck`, `Memory` grows `embed_state`, the web UI shows a distinct chip and Health row, and `memini doctor` counts stuck repairs **and treats them as a warning**; it previously printed "No problems detected" and exited 0 over a permanently keyword-only corpus. Four Prometheus collectors cover the queue.

## What is deliberately not here

`MEMINI_WRITE_EMBED_TIMEOUT` keeps its 5s default. Tightening it was the original plan and it was wrong: before the enrichment replay above, every write pushed onto the degraded path lost that enrichment permanently, and there is no embed-latency data in this repo to justify a specific number anyway. The existing `memini_embed_duration_seconds` histogram is the thing to read before picking one. Now that a degraded write is non-destructive, lowering it later trades write latency for repair latency rather than for quality.

Also recorded in the code, with reasons: `Remember`'s load of an existing row stays fatal (degrading it would reset `CreatedAt`, validity and confidence, which is silent corruption of a row the caller asked to *edit*), the `Forget`/`Supersede` pre-delete snapshot stays inline (moving it off the request path means racing the delete it exists to precede), and consolidation queued by a *healthy* write stays on its lossy channel because there is no durable state from which "needs consolidation" can be re-derived.

## Compatibility

Additive. New columns default to `''` (healthy), so every existing row migrates for free and an older binary ignores them. `metadata.pending_embed` keeps being written for one release so a rollback still finds and repairs the same rows via its own marker scan; the sweeper adopts legacy rows that carry only the marker. Rollback is clean at every point.

## Verification

`mise run test`, `mise run lint`, and all four drift gates (`docs-check`, `rest-check`, `gen-api-check`, `client-check`) pass.

Beyond unit and store-conformance coverage, this was run end to end against a real server with a deliberately broken embedder:

- a degraded save returns with `embed_state: "pending"` and is claimable immediately
- **client-disconnect durability**: `curl --max-time 0.05` aborts the request, and the memory is still there afterwards. That write was lost before this change
- killing the server mid-repair and restarting against a working embedder drains the queue and restores the vector, confirmed by a semantic recall with no keyword overlap
- `/healthz?verbose=1` and `/v1/stats` report the backlog and then zero

That end-to-end pass is also what caught the last bug in here: after a full recovery, `metadata.pending_embed` was still set, so stats/doctor/the UI kept reporting a repaired memory as degraded **and** the sweeper re-adopted it every tick: an infinite repair loop. Fixed in the same statement that clears the columns, and pinned by a conformance subtest so both backends are held to it.

Postgres paths (`FOR UPDATE SKIP LOCKED`, the SQL-computed lease) compile and are covered by the shared conformance suite but have not been run against a live vchord instance here; `mise run test-integration` is worth a pass before merge.

Read commit-by-commit: the seven commits are sequenced so each one builds and tests green on its own.


eleboucher:
> _Originally by **eleboucher** on 2026-08-02 18:52 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/190#issuecomment-14596)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #191), which is the source of truth. Further review and merging happen there.

PR #77, opened 2026-08-04
feat: LLM-assessed intrinsic importance (assessed_importance) with rerank-pool reservation

perfectra1n:
## Problem

`importance` today is either user-supplied or a static tier seed (0.6/0.3/0.1), and access-count reinforcement measures how often a memory *has been* used — nothing assesses whether content is intrinsically important ("root cause of prod outage" vs "user likes tabs"). Meanwhile rerank-pool membership in `finalizeRecall` is decided purely by the composite score, whose production weights zero the standalone importance term — so a genuinely critical memory ranked below the pool cut never even reaches the reranker.

## What this adds

**A separate, nullable `assessed_importance` column** (LLM self-assessed, clamped to `[0.1, 0.9]`) — deliberately not the existing `importance` field, to stay clear of its `!= 0` sentinel. Write-time invariant, enforced and tested: an explicit user-supplied importance always clears the assessment, quarantined writes are never assessed, and `EffectiveImportance()` (assessed if present, else stored) is what ranking reads.

**Three ways the value is produced:**
- Piggybacked on the distill path (`llm.Fact.importance`) and the consolidate path (`llm.Decision.importance`) — same calls, a few extra prompt tokens, nil when the model omits the key.
- A maintenance backfill sweep (`MEMINI_ASSESS_INTERVAL`, default `1h`, only runs when an LLM is configured) that assesses durable rows still at their tier-seed importance — it never second-guesses an explicitly set value.

**Two places it acts, separately gated:**
- **Rerank-pool slot reservation** (`MEMINI_RECALL_IMPORTANCE_RESERVE`, default `2`; `MEMINI_RECALL_IMPORTANCE_MIN`, default `0.75`): reserves pool slots for high-effective-importance candidates that fused retrieval left below the pool cut. Membership-only **by construction** — the `[0,k)` prefix of the deduped list is untouched, so the no-reranker path and every rerank-failure/empty fallback return byte-identical results, and a relevance gate (reusing the durable-reserve ratio/anchor) stops important-but-irrelevant crowding. Structurally inert unless a reranker is configured **and** `MEMINI_RERANK_POOL` exceeds the recall limit.
- **Salience blend** (`MEMINI_ASSESSED_SALIENCE_WEIGHT`, default `0` = exact no-op): opt-in blend of the assessment into `Salience()`. Default-off because it is a ranking *and* lifecycle knob (recall quality term, short-term cap eviction, briefing order, dedup representative).

## On-by-default behavior changes (deliberate — flag if you'd rather flip them)

1. **Backfill sweep** runs hourly on deployments that have an LLM configured (`MEMINI_ASSESS_INTERVAL=0` disables). Cost is bounded: ≤ `MAX_PER_RUN/BATCH` = 10 LLM calls per pass, only while a backlog exists.
2. **Pool reservation** is active (reserve 2) on deployments that already opted into a reranker + widened `MEMINI_RERANK_POOL` — exactly the audience the feature serves. `MEMINI_RECALL_IMPORTANCE_RESERVE=0` disables. Everyone else is bit-identical to today (verified: full suite + bench run without reranker/LLM are unchanged).

## Known limitations (documented)

- The sweep's first-batch-error-aborts + oldest-first-retry semantics mean a persistently unratable head row (e.g. a poison-pill item) can stall the backlog; a skip-after-N/age-out is a natural follow-up.
- In the default async consolidate mode, new/supersede writes are assessed by the sweep (not the consolidator) — update actions stamp directly.
- `DemoteStale`'s `importance >= 0.75` immunity bar deliberately still reads **raw** importance: an assessed 0.9 does not confer demote immunity. Switching it to `EffectiveImportance()` would silently widen immunity across backfilled corpora, so it's left as an explicit follow-up (commented at the site).
- "Explicitly set" on existing rows is inferred as `importance != tier seed`; a caller who explicitly sets exactly the seed value is indistinguishable from untouched (fail-closed on the write path, harmless on the sweep — the number is identical).

## Notes

- Schema change is a backward-compatible nullable column with `ADD COLUMN IF NOT EXISTS` migration in both stores, mirroring `confidence`; the setter deliberately does not bump `updated_at` (assessment is a system annotation and must not reset confidence decay or demote eligibility). Conformance-tested on sqlitevec and real Postgres.
- API exposure is read-only (`readOnly: true` in OpenAPI, REST + MCP outputs); no write surface accepts it.
- New guide section in `docs/guides/tuning-recall.md`; generated reference docs in sync (`mise run docs` clean).
- Bench: `bench/importance_reserve_test.go` sweeps reserve × threshold on a buried-durables corpus (needs a live embedder + reranker to produce numbers).

Commits are structured for review in order: schema → LLM piggyback (+ invariant fixes) → salience knob → pool reservation → sweep → docs → final polish.


eleboucher:
> _Originally by **eleboucher** on 2026-08-04 20:09 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/198#issuecomment-14889)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #199), which is the source of truth. Further review and merging happen there.

PR #78, opened 2026-08-05
feat: ask models to set importance deliberately, and run the assess sweep on demand

perfectra1n:
## Problem

`importance` is the one write-time signal a model sets that ranking never second-guesses. An explicit value permanently clears `assessed_importance`, and the backfill sweep skips any memory whose importance differs from its tier seed — so a value the writing agent chooses is authoritative and final.

**Nothing tells models that.** `serverInstructions` documents `pinned` and `metadata.category` but not `importance`; the `memory_remember` field description explains the mechanics without naming a bar; and `plugin/skills/remember/SKILL.md` documents every other write-time field — `tier`, `tags`, `metadata.category`, `visibility`, `summary`, `confidence` — and omits this one.

The result is a corpus sitting at the tier seed. Measured against a live store: **~90% of durable memories are untouched at 0.6**, and the ~10% clearing the 0.75 reserve bar are the cases where an agent happened to set it unprompted.

That matters more since #199. `MEMINI_RECALL_IMPORTANCE_RESERVE` reads `EffectiveImportance()` — assessed if present, else stored — so on a corpus with no assessments and no explicit values the reserve is dormant by construction. It has nothing to promote.

## Why the write path rather than the sweep

The sweep (`MEMINI_ASSESS_INTERVAL`) is LLM-gated: `cmd/memini/root.go:444` starts it only when a chat client exists. Deployments that deliberately run without an LLM never populate the column at all.

But those deployments' memories are still *authored* by frontier models through `memory_remember`. The writing agent has already read the content — it wrote it — so an importance it sets costs no extra inference, beats a quantized local model's later read of the same text, and outranks any assessment permanently. The signal was already available; it just was never asked for.

An embedder cannot substitute. An embedding encodes what text is *about*, and cosine similarity is relational — it needs two vectors. Importance is a unary, query-independent judgement with no direction in embedding space to project onto. The tempting proxy, distance from the corpus centroid, measures *novelty*: a garbled quarantined write is maximally novel and minimally important, so it anti-correlates exactly where it costs most. The reranker already covers the query-relative axis and access-count reinforcement covers the retrospective one; this is the prospective axis neither expresses.

### 1. Prompt surfaces get a bar, not a rating habit

Calibration is the entire risk, so the guidance names a **threshold with consequences** rather than encouraging models to rate things. `0.75` reserves scarce reranker-pool slots *and* confers immunity from staleness demotion (`demote.go:58`), so the text says: omit by default (the tier seed is correct), spend it on a standing directive, a hard constraint, a bug's root cause, or a user correction — and states plainly that **inflating it is worse than omitting it**, because the reserved band is small and fixed.

That asymmetry is deliberate and worth being explicit about: `clampAssessedImportance` bounds the LLM path to `[0.1, 0.9]` precisely so an over-eager model cannot flatten the scale. **Explicit importance has no such clamp**, because it is the authoritative path. A vague "set importance when it matters" nudge would therefore be worse than silence — it would flood the bar and promote noise into slots that previously held organic candidates.

Placed per the existing architecture rather than everywhere:

- **`serverInstructions`** — the canonical policy MCP hosts carry every turn. Added as a Conventions-adjacent bullet: like `pinned` and `metadata.category`, this is a write-time *convention*, not one of the six mirrored save-policy invariants, so it needs no mirror across the five surfaces named in that file's header comment.
- **The `memory_remember` field description** — what a model reads while filling the value.
- **`plugin/skills/remember/SKILL.md`** — the file that documents every other write-time field.
- **`MEMORY_INSTRUCTION` deliberately untouched.** It is the abridged trigger with a documented ~700-char budget and is already at 698. Its own comment says to let the server text be the single source of truth; adding here would violate the budget and duplicate.

### 2. `memini assess` — the sweep, on demand

Turning the timer off shouldn't mean giving up the column. This exposes the same pass as a command, so `MEMINI_ASSESS_INTERVAL=0` can stay set permanently and `MEMINI_LLM_BASE_URL` can point at a model for the duration of one invocation instead of the life of the process.

It calls `AssessImportanceBackfill` directly, so an on-demand run and a timed one are calibrated identically **by construction** — there is no second prompt to drift.

**Dry-run by default**, matching `memini backfill`:

```
$ memini assess
assess importance: preview (re-run with --yes to apply)
  unassessed at tier seed: 0 durable memories
  would cost:              0 LLM call(s) at batch 20
```

The preview is the reason for the new `AssessImportancePreview`. The cost of this pass is per-row LLM spend, so an operator needs the size of the bill before agreeing to it — and a preview that called the model to produce that number would defeat itself. It shares `assessCandidates` with the real pass, so the count is **exact rather than estimated**, and its signature takes no assessor, so a preview cannot spend a call even by mistake.

Two smaller deliberate choices: `--yes` without an LLM fails with a clear message rather than silently assessing nothing, and the applied path prints the count *before* returning any error, because the pass persists per batch and a mid-run failure still leaves committed work the operator needs to know about.

## Testing

`TestAssessImportancePreviewMatchesBackfill` pins the invariant the preview exists for — that its count matches what the backfill actually assesses, before and after a run — guarding against the preview growing its own selection logic and quoting a cost that doesn't match the bill.

Verified: full suite green, `golangci-lint` 0 issues, `mise run docs-check` clean. Smoke-tested both CLI paths (dry-run, and the `--yes` no-LLM guard).

## Notes

- No behavior changes to any existing code path — the prompt edits are text, and `memini assess` is a new command.
- Generated reference docs regenerated; `docs/guides/tuning-recall.md` now points at the command where it previously only documented turning the timer off.
- Follows #199, which added the reserve this guidance feeds.


eleboucher:
> _Originally by **eleboucher** on 2026-08-13 09:06 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/200#issuecomment-15472)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #209), which is the source of truth. Further review and merging happen there.

PR #79, opened 2026-08-17
fix: activity-log inject rows stored bare when beacon hydration races its serve

perfectra1n:
## What this looks like

The Recent Activity feed occasionally shows `inject` events whose memory rows render as a bare id prefix (`#1 c1e80069`) with no summary — and clicking such a row throws a 404, because the UI has no namespace to open the memory with and falls back to the viewer's active one.

## Why it happens: the append "queue" doesn't exist

`RecordInjected` hydrates the beacon's bare ids by *borrowing* the snapshot from the serve (recall/briefing) event that preceded it — `hydrateInjected` reads `ServedSnapshots` for the same namespace. The comment on `logEventsPrepared` asserted this was safe because the beacon's background write is "already queued behind the serve that preceded it".

No such queue exists. `s.bg` is a plain `sync.WaitGroup`; `bg.Go` spawns free-running goroutines that the scheduler may interleave arbitrarily. The failing sequence:

1. A hook calls recall → the serve rows are appended from **goroutine A**; the response returns immediately.
2. The plugin injects and POSTs the beacon within milliseconds.
3. **Goroutine B** (the beacon) runs `hydrateInjected`, which reads for A's rows.
4. Under store write contention, A's insert hasn't committed when B's read runs → hydration finds nothing → the inject rows are stored **permanently bare**.

Observed on a busy sqlite deployment (~1.4 GB db): **~3% of injected rows stored bare** over a 15h window, with serve→beacon gaps of 7–21 ms. The definitive fingerprint: for every bare row, a matching serve event exists in the *same namespace*, timestamped milliseconds *before* the beacon — and byte-identical event sequences with the same ~8 ms gap hydrate fine on other occasions. Same namespace, serve always present, outcome flips on timing. (Two other theories — the 24h `injectHydrateWindow` and a namespace mismatch — were both falsified by this data.)

## The fix, in three layers

**1. Make the documented ordering real (prevention).** Async activity-log appends are now chained FIFO: under a mutex, each append captures the previous append's `done` channel as its predecessor and installs its own as the new tail; the goroutine waits on its predecessor before writing. Submission order is now commit order, so the beacon's hydration read always observes the serve that preceded it.

Design notes:
- The mutex guards only the handoff (capture prev / install self), never the I/O — no serialization cost on the request path, and `logEventsPrepared` still returns immediately.
- The per-write `eventLogTimeout` starts *after* the predecessor wait, so a slow append consumes its own budget rather than cascading deadline exhaustion down the chain.
- No resident worker goroutine, no close/drain lifecycle: `WaitBackground()` already joins the last link via the existing WaitGroup. `WithSyncEventLog` (tests) is unchanged.

**2. Heal historical rows at read time (repair).** Logs written before this fix carry bare rows sprinkled through history. `Events()` now re-runs hydration over still-bare inject rows when assembling a feed page — grouped by each row's own event namespace (a page can span namespaces, and the borrow-from-serve security rule must hold per row), unbounded in time (a historical row's serve sits near the *row's* moment, not near now), and display-only: nothing is written back, so the "ids are taken on faith, never resolved against the memories table" containment is untouched. A row that still resolves nothing stays bare. Known limit (deliberate): still-bare rows continue to miss the store-side tier/text filters; only a persisted backfill could change that.

**3. Render the truly unresolvable inert (UI).** A memory row with neither summary nor namespace can't be opened — a click could only 404. It now renders as a dimmed, non-interactive row with a `no snapshot` badge and an explanatory tooltip, instead of a button that promises a click it cannot honor.

## Tests

- `TestAsyncEventLogOrdersServeBeforeBeacon` — regression for the race. Runs the event log **async** (the production default) against a store wrapper that delays recall-row appends by 100 ms, then asserts on the **raw stored rows**. Verified to fail against the pre-fix code with exactly the production symptom (`MemoryNS:"" MemoryTier:"" MemorySummary:""`) and pass with the FIFO. It bypasses `Events()` on purpose: the read-time heal would otherwise mask a regression in the prevention layer.
- `TestEventsHealsBareInjectRows` — appends a pre-fix-style bare row directly, asserts `Events()` returns it hydrated from a 48h-old serve, that a never-served id stays bare (containment), and that nothing was written back to the store.
- Full `go test ./...`, `golangci-lint` (0 issues), and UI `tsc`/build pass.

## Compatibility

No API, schema, or store-interface changes. `hydrateInjected` gains a `since` parameter (internal); both sqlite and postgres `ServedSnapshots` already accept a zero bound.

### DIFF
--- a/internal/service/events.go
+++ b/internal/service/events.go
@@ -89,10 +89,19 @@ func (s *Service) logEvents(ctx context.Context, events []store.Event) {
 // logEventsPrepared is logEvents with a hook that runs against the rows just
 // before they are appended, for a writer that must read the store to complete
 // them (RecordInjected's snapshot hydration). Running inside the same goroutine
-// as the append keeps the read off the caller's latency, and — since the
-// background write is already queued behind the serve that preceded it — gives
-// that serve's own rows a head start on landing, which is exactly what the
-// hydration looks for. prepare must be best-effort: it cannot fail the append.
+// as the append keeps the read off the caller's latency, and the FIFO chain
+// below guarantees the serve that preceded the beacon has COMMITTED before the
+// hydration read runs, which is exactly what it looks for. prepare must be
+// best-effort: it cannot fail the append.
+//
+// The FIFO is load-bearing, not an optimization. An earlier version spawned
+// each append as a free-running goroutine and asserted in a comment that the
+// beacon's write was "already queued behind the serve that preceded it" — no
+// such queue existed (sync.WaitGroup orders nothing), and under store write
+// contention the beacon's hydration read raced the serve's insert and lost,
+// permanently storing the inject rows bare (~3% of injections on a busy
+// sqlite deployment, observed 7-21ms serve→beacon gaps). Chaining each append
+// on its predecessor's completion makes submission order the commit order.
 func (s *Service) logEventsPrepared(ctx context.Context, events []store.Event,
 	prepare func(context.Context, []store.Event),
 ) {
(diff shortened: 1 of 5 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-08-17 18:04 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/218#issuecomment-15756)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #221), which is the source of truth. Further review and merging happen there.

PR #80, opened 2026-08-17
docs: how-it-works layer, test-pinned examples, deployment expansion; fix stored:false tier + MCP namespace normalization

perfectra1n:
## Why

The docs answered "which knob do I turn" well (guides, generated reference) but not "how does this thing actually work" or "how do I run it for real". Concretely missing: when and how namespaces come into existence, how a write decides where to land and what tier it gets, when the plugin pulls memories into context, and the whole story of obtaining/deploying released artifacts. On top of that, an audit turned up documentation that was actively wrong, plus two real behavior bugs the audit surfaced.

## What this adds

**`docs/how-it-works/` — five mechanism pages plus an overview** (`write-path`, `recall`, `namespaces`, `lifecycle`, `plugin`). Operator-level concept docs, no file paths in the body; each ends with a `Source map` section so contributors can jump to the code. Every mechanism was verified against source while writing — the tier-classification examples were run through the actual classifier, not guessed.

**`docs/examples/` — five end-to-end walkthroughs, each pinned by a Go test.** Every quoted payload/response field is asserted by a dedicated example test against the real in-process stack (`internal/service/example_*_test.go`, `internal/api/rest/example_*_test.go`), and each doc carries a `Validated by` footer naming its tests — the walkthroughs cannot silently drift from the code:
- `first-memory` — fresh deploy to a remembered fact (namespace materializes on first write)
- `namespace-lifecycle` — lexical ancestors, visibility routing up the tree, subtree reads, pins, disappearance
- `memory-life-story` — one fact through classification, reinforcement, the value gate, merge hints, supersession, `as_of` time travel, and history
- `recall-in-practice` — one query at scope `project`/`full`/`everywhere`, plus the exact briefing scope header and the briefing-never-reinforces asymmetry
- `team-sharing` — two named keys, personal vs team writes, a read-only 403, and links narrowing (never widening) a read set

**Operations expansion.** `deployment.md` now documents the published registries, tags, OCI Helm chart, signed tarballs, and `cosign verify` recipes (sourced from `release.yml`, nothing invented), a fourth systemd shape (with the cwd-relative `MEMINI_SQLITE_PATH` trap defused), and the chart-from-checkout tag trap (verified: `helm template` from a checkout resolves the placeholder `0.1.0` tag → ImagePullBackOff). New `production.md` (TLS/reverse proxy + the `MEMINI_REQUIRE_HTTPS` interplay, sizing, Postgres/VectorChord operations, keys-file has no live reload, metrics topology) and `backup-restore.md` (per-backend hot backups, and the embed-model/dims restore invariants). New `CONTRIBUTING.md` documents the dev loop and every generated-code drift gate, including the vendored-deps requirement for `helm-docs`.

`docs/scopes.md` gains an appended `Fine print` section (read-set caps, origin-once, the lexical-ancestors vs subtree asymmetry, widest-tiers-wins); the `## Knobs` heading and `#knobs` anchor cited by the boot-fatal config errors are byte-untouched (the pin test still passes).

## Behavior fixes (both test-first)

1. **`stored:false` now reports the tier the write resolved to.** Both surfaces echoed the raw input tier on a value-gated write — `""` whenever the agent omitted it, so the agent learned nothing. `Remember` gains an `EffectiveTier` out-param (same pattern as `MergeHint`/`Reinforced`), populated as soon as validation succeeds so it survives the gate drop; MCP and REST report it, and the previously-undocumented `200 stored=false` response is now in the OpenAPI spec.
2. **MCP normalizes `X-Memini-Namespace` like REST.** MCP captured the header TrimSpace-only while REST canonicalized it, so `"team/proj/"` addressed two different namespaces depending on transport (acknowledged in a code comment as deferred). Both MCP capture sites now use `httputil.NormalizeNamespace`, with a transport-equivalence test. Migration: rows written over MCP under a non-canonical header live in a sibling namespace — `memini doctor` now flags every non-canonical stored namespace and prints the `memini namespace move` command, and `upgrading.md` documents the path. The official plugin is unaffected (it resolves namespaces via the handshake, which always produced canonical names).

## Documentation defects fixed

- `README.md` told Codex users to run `memini serve` — no such subcommand exists (the server is the bare `memini` root command).
- `deployment.md` claimed remember/recall "error out" without an embeddings endpoint. Verified false under default timeouts: writes store vectorless (`pending_embed`, background repair) and recall degrades to keyword-only. The boot warning in `cmd/memini/root.go` and the `embed.Disabled` doc comment carried the same stale claim and are fixed alongside.
- `README.md` claimed `MEMINI_URL`/`MEMINI_TOKEN` "remain legacy aliases" — they are removed and warn-once-ignored.
- Hook/skill counts drifted across three manifests ("6 hooks" / "7 hooks" / "3 skills" vs the 7 hooks, 8 commands, 9 skills actually shipped).
- The chart README said metrics live at `:8080/metrics`; the chart's own default is a dedicated `:9090` port (fixed in the gotmpl and regenerated).
- Deleted `docs/dashboards/memini-grafana.json` — an orphaned, diverging copy of the chart's dashboard, referenced by nothing — and **rebuilt the canonical chart-shipped dashboard** (`charts/memini/dashboards/memini.json`): the old one covered 24 of 43 registered metrics and had real query defects (a `label_replace` whose synthesized label was then discarded, tokens+errors sharing one axis at incompatible scales, a dedup ratio diluted by gated writes, multi-replica double counting). Now 8 rows / 46 panels, 43/43 metrics covered — including the previously-invisible degraded writes and the self-repair backlog (with threshold stat tiles in the Now row) — every query machine-verified against the metric registry, and every panel carrying a what-bad-looks-like description. Two instrumentation gaps found along the way are fixed: briefings now record `op_duration` (they were invisible server-side), and demotions get a new `memini_demoted_total` counter via a `DemoteStale` report hook (they were log-only). Remaining uninstrumented signals (rerank duration histogram, repair-age distribution, consolidation queue wait, embed cache hit ratio) are listed as follow-ups.
- `MEMINI_INJECT_TELEMETRY` existed in code but no README table.
- Dockerfile OCI label said `MIT`; the repo is AGPL-3.0. Also aligned the two plugin manifests that said `Apache-2.0`. **Maintainer call:** if the plugin was meant to be more permissively licensed than the server, revert `plugin/.claude-plugin/plugin.json` / `plugin/.codex-plugin/plugin.json` — the packages published to npm (`@memini/client`, opencode) still say MIT and were deliberately left alone for the same reason.

## Flagged, not fixed

- **The cosign public key is not committed or published anywhere** — the new verification docs can only say "obtain it out of band". Committing a `cosign.pub` (or switching to keyless) would complete the supply-chain story.
- Codex's MCP namespace is static (`env_http_headers`) — per-project resolution works for the hooks but not the MCP tools under Codex. Now documented as a known limitation; a code fix needs upstream Codex support.
- The client-side env surface has no CI drift gate (the server side does, and it is clean).

Formatting: all markdown this branch touches is reflowed to one line per paragraph (no 80-column hard wrapping); oxfmt's proseWrap-preserve keeps it that way.

## Verification

All suites and gates green locally: `mise run test`, `test-hooks`, `test-client`, `docs-check`, `client-check`, `rest-check`, `gen-api-check`, `helm-docs-check`. New tests: `TestExampleMemoryLifeStory`, `TestExampleNamespaceLifecycle`, `TestExampleRecallScopes`, `TestExampleTeamSharing`, `TestExampleFirstMemory`, `TestRememberEffectiveTierOutParam`, `TestRememberToolDroppedWriteReportsEffectiveTier`, `TestRememberLowSignalReportsEffectiveTier`, `TestHTTPHandlerNormalizesNamespaceHeader`, `TestWarnNonCanonicalNamespaces`. Every relative link in the docs tree resolves; the `docs/scopes.md#knobs` pin test passes.


eleboucher:
> _Originally by **eleboucher** on 2026-08-17 18:04 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/219#issuecomment-15757)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #220), which is the source of truth. Further review and merging happen there.

PR #81, opened 2026-08-21
fix(plugin): restore MCP auth on Claude Code >= 2.1.238 via 0600 credentials-file fallback

perfectra1n:
## Summary

Claude Code **2.1.238** stopped passing inherited credential env vars to `headersHelper` scripts registered by a plugin (intentional hardening, per its changelog). memini's helper read the bearer exclusively from `MEMINI_API_KEY`, so on 2.1.238+ it emitted only the namespace header, the server 401'd, Claude Code fell into OAuth discovery, and the connection died with `Dynamic Client Registration rejected (HTTP 404)` — every memini MCP tool silently vanished while the hooks (full env) kept working. Reads looked healthy; writes were impossible.

This PR restores auth without depending on env inheritance:

- **Fix A — env-independent credential source** (`packages/memini-client/src/credential.ts`): the SessionStart hook (full env) mirrors `MEMINI_API_KEY` into `${XDG_CONFIG_HOME:-~/.config}/memini/credentials` (mode `0600`, dir `0700`, atomic write, entries keyed by normalized base URL so a bearer stored for one server is never sent to another). `mcp-headers.mjs` falls back to it when its env carries no key; an env key still wins, and unsetting `MEMINI_API_KEY` retires the stored copy at the next session start. Deliberately **not** gated on a `MEMINI_API_KEY_FILE`-style env var — any `KEY`/`TOKEN`/`SECRET` name would be stripped by the same heuristic.
- **Fix C — loud, condition-based detection** (no version sniffing): the helper warns on stderr (→ Claude Code's MCP logs) when the session's cached handshake proves the hooks authenticated (`identity.authenticated`) but the helper ended up with no bearer from env or file; SessionStart warns when the credential write fails while a key exists. The prior failure mode's worst property was silence.
- **Fix B — documented user-scope workaround** for stale plugin versions (`plugin/README.md`): `claude mcp add --scope user ...` reusing the plugin's own helper via the plugin-root breadcrumb, with the duplicate-toolset and stale-✘ caveats.
- Docs: `docs/reference/env-vars.md` client `MEMINI_API_KEY` row; `.mcp.json` comment updated (its inline `printf` env fallback only helps pre-2.1.238 — it fires only before any hook has run, when the credentials file can't exist yet either).

## Security tradeoff (stated, not glossed)

This puts a bearer at rest on disk where any process running as the user can read it — a genuine, partial re-opening of what the upstream hardening closed. Mitigations: memini's own file at `0600`, holding only memini's key; base-URL keying prevents cross-server bearer travel; the stored copy is retired when the env var goes away. The hardening targeted inheritance *breadth* (every plugin seeing every `*_API_KEY`), not secrets-at-rest; users who reject the tradeoff can use the documented user-scope registration instead.

## Testing

- `packages/memini-client/test/credential.test.ts`: path resolution, roundtrip, `0600`/`0700` modes, corrupt-file / base-URL-mismatch misses, retire-on-empty, env-wins fallback (187/187 client tests pass).
- `plugin/scripts/_test.mjs`: helper falls back to the stored credential with `MEMINI_API_KEY` stripped from the env (the exact 2.1.238 repro), ignores a mismatched base URL, warns loudly when the cached handshake shows auth expected but no key; SessionStart mirrors and retires the file, `0600` asserted (223/223 hook tests pass).
- `mise run client-check` (bundle drift gate), `mise run docs-check`, full Go suite — all green.
- Direct repro: `syncStoredApiKey(...)` then the helper under `env -u MEMINI_API_KEY` emits both headers again (`X-Memini-Namespace` + `Authorization`).

Out of scope, per the diagnosis: server-side OAuth/DCR (huge protocol surface to avoid sending one header the plugin already has) and the upstream Anthropic ask for a consented env-var allowlist (long game; not blocked on).


eleboucher:
> _Originally by **eleboucher** on 2026-08-22 11:44 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/222#issuecomment-16074)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #223), which is the source of truth. Further review and merging happen there.

PR #85, opened 2026-08-31
Eliminate the MCP DCR-404 failure class: stateless HTTP, tri-state credential retirement, self-diagnosing OAuth probes

perfectra1n:
## The problem

A plugin user on v0.7.21 reported the memini MCP server "dies all the time" in Claude Code, with `/mcp` showing:

> Failed to reconnect to plugin:memini:memini: Dynamic Client Registration rejected (HTTP 404): {}

That message is an auth-loss signature, not an OAuth problem. The chain: a reconnect reaches the server without a valid bearer → bare 401 (no `WWW-Authenticate`) → Claude Code assumes the server wants OAuth and runs discovery → `POST /register` hits the SPA fallback's `{}` 404 → the cryptic error above, until the user reloads all plugins.

Two defects made this chronic:

1. **Credential-retirement footgun.** The Claude Code ≥ 2.1.238 mitigation (credentials file written by SessionStart, read by the headersHelper) had a destructive edge: `syncStoredApiKey()` deleted the stored credential whenever a SessionStart ran with `MEMINI_API_KEY` unset — and `readBootstrap` collapses unset and explicitly-empty (`env["MEMINI_API_KEY"] || ""`), so *any* session launched from an env without the var (IDE/GUI launcher, Codex — same hook script) silently broke MCP auth for every other session. "Reload plugins" fixed it only because SessionStart re-mirrored the key — until the next keyless launch. The shipped test suite actually pinned this buggy behavior.
2. **In-memory MCP sessions.** Every server restart 404'd all sessions ("session not found"), forcing reconnects through that fragile auth path. Stateful sessions also froze identity at creation: requests 2..N ignored changed `X-Memini-Namespace`/`X-Memini-Home` headers, and a different valid key reusing a session ID inherited the original session's namespace/home/read-only flag.

## The fix, and why this shape

**1. Stateless streamable HTTP** (`feat(mcp)`). `StreamableHTTPOptions{Stateless: true}` + a shared package-level `SchemaCache` (the SDK documents it for exactly this one-server-per-request deployment; all tool schemas are package-level vars, so the cache stays bounded). memini's MCP surface is tool-only — no sampling/elicitation/subscriptions/notifications — and already derives all identity from request headers inside `getServer`, so statelessness is a strict fit: restarts can no longer invalidate anything, the session leak (go-sdk #499) disappears, and the frozen-identity bug is fixed as a side effect (now pinned by `TestHTTPHandlerPerRequestNamespace` / `TestHTTPHandlerPerRequestReadOnly` / `TestHTTPHandlerStatelessIgnoresStaleSessionID`). Deliberately NOT set: `KeepAlive` (server→client ping, rejected under stateless), `JSONResponse` (kept the default, better-tested SSE framing), `SessionTimeout` (nothing left to expire).

**2. Tri-state credential retirement** (`fix(client)` + `fix(plugin)`). `syncStoredApiKey(baseUrl, apiKey: string | undefined)`: `undefined` → new `"kept"` action, zero I/O (a keyless env carries no opinion and must not destroy what other sessions depend on); explicit `""` → retire, now logged on stderr (the deliberate revocation gesture); non-empty → upsert. The tri-state lives in the library rather than a call-site guard so any future caller gets the safe behavior; `Bootstrap.apiKey` stays `string` to avoid rippling `undefined` through every consumer — the one call site passes the raw env value, with a comment explaining why `boot.apiKey` is a trap. The old test that pinned the deletion was rewritten into the two regression cases (keyless env → file untouched; explicit-empty → retired + stderr), and a failed retirement now warns that the bearer is still on disk instead of failing silently.

**3. Self-diagnosing OAuth-probe responses** (`feat(server)`). Explicit routes for `POST /register` and `GET /.well-known/{oauth-protected-resource,oauth-authorization-server,openid-configuration}` (+ RFC 9728 path-suffixed variants) return 404 with a JSON body whose `error` text Claude Code surfaces verbatim — so the user-visible message *becomes* the diagnosis:

```
{"error":"memini does not use OAuth (no valid bearer reached the server). plugin: run /memini:status, see plugin README 'Claude Code 2.1.238 and credential env vars'; behind a proxy: check it forwards Authorization"}
```

Registered unconditionally on the base router, so all three UI modes (catch-all, dedicated `MEMINI_UI_ADDR`, UI disabled) now answer identically — previously only the SPA catch-all mode returned parseable JSON. The `ui.go` blanket `{}` branch stays as defense in depth. The MCP and `/metrics` 401s now go through `httputil.Error` (they were `http.Error` with a JSON literal → `text/plain` Content-Type) and carry `WWW-Authenticate: Bearer realm="memini"` per RFC 6750, each in a one-line hunk that's trivially revertable.

**4. Proactive drift detection** (`feat(plugin)`). `/memini:status` gains `cred-file-stale` (warn: env key differs from the stored credential the headersHelper will use) and `cred-file-only` (info: no env key here, but MCP still authenticates via the file). Explicit-empty produces neither, by design.

**5. Docs.** `plugin/README.md` documents the tri-state; `docs/operations/production.md` gains "Troubleshooting: `Dynamic Client Registration rejected (HTTP 404)`" (greppable verbatim; covers the reverse-proxy-strips-`Authorization` cause too); `upgrading.md` notes the two visible behavior changes (restart survival, `GET /mcp` → 405 `Allow: POST`); `env-vars.md` caught up.

## Verification

- Full matrix green: `go test ./...`, `golangci-lint` (0 issues), `@memini/client` tests, plugin hook tests (228/228, including the new regression tests), bundle drift check, docs-check.
- HTTP smoke against a built server: probe routes return the self-diagnosing JSON; 401 carries `WWW-Authenticate` + `application/json`; `tools/list` with a bogus `Mcp-Session-Id` succeeds, **including across a full server restart** — the reconnect-storm class is gone.
- Every commit was independently code-reviewed (spec + quality) before landing; a final whole-branch review's three must-fix items are folded in as the last commit.

## Behavior changes to know about

- `GET /mcp` (standalone SSE stream) now returns 405 `Allow: POST` — spec-correct for stateless servers, and memini never pushed anything on that stream, but strict clients may log it. Verified fine at the HTTP level; worth one manual pass with a live Claude Code before release.
- Namespace/home headers and key identity now take effect per request instead of freezing at session creation (the freezing was a bug, but it is observable).
- A rotated-away `MEMINI_API_KEY` no longer scrubs the stored credential on the next session start — set `MEMINI_API_KEY=""` once (or delete `~/.config/memini/credentials`) to retire it deliberately.


eleboucher:
> _Originally by **eleboucher** on 2026-08-31 07:00 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/244#issuecomment-16825)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #245), which is the source of truth. Further review and merging happen there.

PR #88, opened 2026-09-04
feat(chart): optional PrometheusRule with recall-degradation alerts

### LINKED ISSUE #87: Bundle alerting rules in the chart, and export the rerank concurrency limit as a gauge
### Summary

The chart ships a Grafana dashboard but nothing that fires, and memini's two degradation paths are both quiet: they keep serving requests and surface as worse retrieval rather than as errors. On a deployment with low recall volume, a healthy instance and a degraded one are indistinguishable unless something is watching the counters.

Two things would close that gap. I have the first ready as a PR; the second needs a small change in the service and I am happy to do that too if you want it.

### 1. An optional PrometheusRule in the chart

`memini_recall_degraded_total` and `memini_embed_errors_total` are already exported and are enough on their own:

- **recall degraded** — the reranker or embedder was unavailable and results fell back to unranked order
- **embed backend errors** — with a hold time, since these are often a brief blip

Both are pure metric-name predicates with no deployment-specific assumptions, so they seem reasonable to bundle. I have mirrored the `grafanaDashboards` key: off by default, per-rule window/hold/severity, and a `labels` map for whatever selector a Prometheus CR's `ruleSelector` matches on.

### 2. Export the configured rerank concurrency limit

This is the more interesting one, and it is why I am not proposing a saturation alert in the PR.

`memini_rerank_in_flight` is exported but the configured ceiling (`MEMINI_RERANK_MAX_CONCURRENCY`) is not, so there is no portable way to express "rerank has been pinned at its limit". The best I could write for my own deployment hardcodes my own concurrency value, which is not something worth shipping in a chart:

```
min_over_time(memini_rerank_in_flight[15m]) >= 2
```

A gauge for the configured limit, set once at startup, would let anyone express it as a ratio and have it hold across reconfiguration:

```
min_over_time(memini_rerank_in_flight[15m]) / memini_rerank_concurrency_limit >= 1
```

This matters most when the reranker is slow rather than broken. I moved my embedder and reranker onto CPU-only nodes to reclaim memory on the accelerator, which took a realistic 8-candidate rerank from about 1.3s to about 4s. That is fine for my volume, but it is exactly the regime where in-flight sits at the ceiling for long stretches without anything erroring, and there is currently no way to alert on it without baking in a local constant.

Same argument would apply to embed concurrency if that has a configured ceiling.

joryirving:
Closes #87 (the chart half of it).

The chart already ships a Grafana dashboard but nothing that fires, and both of memini's degradation counters describe quiet failures — recall keeps serving when the reranker or embedder is unavailable and just falls back to unranked order, and a failing embed backend shows up as worse retrieval rather than an error — so on a deployment with low recall volume a healthy instance and a degraded one look identical unless something watches the counters. This adds a `prometheusRule` key mirroring `grafanaDashboards`: off by default, two rules over `memini_recall_degraded_total` and `memini_embed_errors_total`, per-rule window/hold/severity, and a `labels` map for whatever selector a Prometheus CR's `ruleSelector` matches on.

Deliberately omitted the rerank-saturation rule discussed in #87, because it cannot be written portably until the configured concurrency limit is exported as a gauge.

Checks run: `mise run helm-docs` (README.md + values.schema.json regenerated, idempotent on re-run, deps vendored first per CONTRIBUTING), `mise run helm-lint` clean, and `helm template` verified to render nothing by default and a valid PrometheusRule when enabled. The rule expressions were checked against a live Prometheus that has these metrics.


eleboucher:
> _Originally by **eleboucher** on 2026-09-10 07:17 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/247#issuecomment-17515)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #270), which is the source of truth. Further review and merging happen there.

PR #89, opened 2026-09-07
Fix possible fix(deps): 5 vulnerable dependencies in go.mod

begininvoke:
This changes `go.mod` to address something a scan flagged. It is around line 1.

The ValidationHandler in kin-openapi incorrectly replaces a nil AuthenticationFunc with NoopAuthenticationFunc, which always returns nil and bypasses any security checks. This creates a fail-open path: omitted AuthenticationFunc results in all security requirements being satisfied, allowing unauthenticated attackers to access protected routes. The risk is critical because typical usage omits AuthenticationFunc, leading to an authentication bypass vulnerability (GHSA‑r277‑6w6q‑xmqw).

Update vulnerable dependencies to fixed versions.

For reference: rule `GHSA-r277-6w6q-xmqw`. Rated critical.

I may well be missing context here — if the current code is deliberate, feel free to close this.

---
*Found with automated scanning ([RedGem](https://code.redgem.net)) and reviewed before opening. If it is not useful, closing it is completely fine.*

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,85 +3,85 @@ module github.com/eleboucher/memini
 go 1.26.5
 
 require (
-	github.com/asg017/sqlite-vec-go-bindings v0.1.6
-	github.com/go-chi/chi/v5 v5.3.0
-	// Pinned: newest ncruces with both sqlite3.Binary and wasm atomics, required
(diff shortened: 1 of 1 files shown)


eleboucher:
updated with renovate

PR #94, opened 2026-09-07
fix(openclaw): pass the configured timeout when registering hooks

### LINKED ISSUE #93: OpenClaw plugin: timeout_ms never applies to the recall hook (capped at OpenClaw's 15s default), while capture keeps its 30s
The plugin registers its hooks without options:

```ts
// integrations/openclaw/plugin/src/index.ts:2047
const addHook = (name: string, handler: any) => {
  if (typeof api.on === "function") api.on(name, handler);
  ...
};
addHook("before_prompt_build", recallHandler);   // :2051
addHook("agent_end", captureHandler);            // :2089
```

OpenClaw resolves a typed hooks timeout as:

```js
// loader: resolveTypedHookTimeoutMs
policy?.timeouts?.[hookName] ?? policy?.timeoutMs ?? opts?.timeoutMs
```

falling back to `DEFAULT_MODIFYING_HOOK_TIMEOUT_MS_BY_HOOK`, where
`before_prompt_build` is **15000ms**. Since the plugin passes no `opts`, the
configured `timeout_ms` (default 30000, and settable far higher) never reaches the
recall hook. `api.on` does forward a third argument:

```js
on: (hookName, handler, opts) => registerTypedHook(record, hookName, handler, opts, params.hookPolicy)
```

## Why it matters

The two hooks land on different budgets. `agent_end` is a void hook with a 30s
default, so **capture gets 30s while recall gets 15s**. When the backend is slow,
recall is killed and discarded but capture still writes — retrieval degrades while
the store keeps growing, which is the worst way for it to fail.

Observed with a CPU cross-encoder reranker: a single recall over the default
8-document pool measured 12.0s end to end with nothing else in flight, leaving ~3s
of margin. Any variance produced:

```
[hooks] before_prompt_build handler from memini failed: timed out after 15000ms
```

repeatedly, while `POST /v1/memories` continued at one write per turn. Raising the
plugin `timeout_ms` to 60000 changed nothing, which is what led to this.

## Workaround

Setting the timeout in OpenClaw config rather than plugin config works, because
`policy.timeouts` wins over everything:

```jsonc
plugins: { entries: { memini: { hooks: { timeouts: { before_prompt_build: 60000 } } } } }
```

This is not mentioned in the plugin docs, and it is surprising that the plugin`s
own `timeout_ms` is the one knob that does not work here.

## Suggested fix

Pass the resolved timeout at registration so `timeout_ms` acts as the default for
both hooks, while still letting `policy.timeouts` override it. `sessionCtx.cfg` is
already in scope at the registration site. PR to follow.

joryirving:
Fixes #93.

`api.on` forwards a third options argument:

```js
on: (hookName, handler, opts) => registerTypedHook(record, hookName, handler, opts, params.hookPolicy)
```

and OpenClaw resolves the timeout as `policy.timeouts[hookName] ?? policy.timeoutMs ?? opts.timeoutMs`.
The plugin passed no options, so both hooks used OpenClaw`s per-hook defaults —
15s for `before_prompt_build`, 30s for `agent_end` — and `timeout_ms` never applied
to recall at all.

The asymmetry is the damaging part: with a slow backend, recall is killed at 15s
and discarded while capture still writes at one memory per turn, so the store grows
while retrieval degrades. Measured against a CPU cross-encoder reranker, a single
recall over the default 8-document pool took 12.0s with nothing else in flight,
leaving ~3s of margin before the cap.

This passes the already-resolved `sessionCtx.cfg.timeout_ms` as `opts.timeoutMs` for
both hooks. Because `policy` is checked first, a host that sets
`plugins.entries.memini.hooks.timeouts.before_prompt_build` still wins — this only
changes the fallback, from OpenClaw`s built-in default to the plugin`s own
configured timeout.

`Number(...) || undefined` keeps the current behaviour when `timeout_ms` is unset or
zero, so the option is simply omitted and OpenClaw`s default applies as before.

Build, typecheck and the plugin test suite (72 tests) all pass.

### DIFF
--- a/integrations/openclaw/plugin/src/index.ts
+++ b/integrations/openclaw/plugin/src/index.ts
@@ -2044,11 +2044,17 @@ const plugin: {
     // before_prompt_build/agent_end (the coarse api.registerHook is for internal
     // events like message:sent). Warn rather than silently drop memory if a build
     // somehow lacks it (eleboucher/memini#26).
-    const addHook = (name: string, handler: any) => {
-      if (typeof api.on === "function") api.on(name, handler);
+    // OpenClaw resolves a hook's timeout as
+    // policy.timeouts[hookName] ?? policy.timeoutMs ?? opts.timeoutMs, falling back to
+    // its own per-hook defaults (before_prompt_build: 15s, agent_end: 30s). Passing the
+    // configured timeout as opts makes timeout_ms the default for both hooks while
+    // leaving host config free to override it (eleboucher/memini#93).
+    const addHook = (name: string, handler: any, opts?: { timeoutMs?: number }) => {
+      if (typeof api.on === "function") api.on(name, handler, opts);
       else api.logger.warn?.(`memini: api.on unavailable; ${name} hook not registered`);
     };
-    addHook("before_prompt_build", recallHandler);
+    const hookTimeoutMs = Number(sessionCtx.cfg.timeout_ms) || undefined;
+    addHook("before_prompt_build", recallHandler, { timeoutMs: hookTimeoutMs });
 
     const captureHandler = async (event: any, hookCtx: any) => {
       const live = await sessionLive(sessionCtx);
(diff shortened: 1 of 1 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-09-07 21:12 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/263#issuecomment-17329)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #264), which is the source of truth. Further review and merging happen there.

PR #96, opened 2026-09-08
feat(openclaw): capture_skip_patterns to drop turns matching a marker

joryirving:
The only capture noise filter, `startsWithNoisePrefix`, matches a **prefix of the user text**. Integrations that emit a structured block mid-reply slip past it. Concretely: an image-generation tool invocation puts `type: image generation task` in the assistant output, the turn is captured verbatim, then promoted into a durable semantic "fact" that recall keeps surfacing. On one deployment this class reached ~29 memories, one recalled 87 times.

Adds `capture_skip_patterns`: a list of case-insensitive substrings matched across the **combined user + assistant** turn; any match skips capture. Empty by default, so existing deployments are unaffected. An operator sets e.g. `["type: image generation task"]` and those turns never enter the corpus.

It generalises `NOISE_PREFIXES` (prefix-of-user-text only) to an operator-configurable substring backstop that matches anywhere in the turn, which is where these markers actually land.

Wired at every site a list-config needs, mirroring `system_kinds`: the TypeBox config schema, `resolveConfig` (with an empty-default fallback), the new `matchesSkipPattern` helper, the `agent_end` capture handler, and the manifest `configSchema` allowlist -- a key missing there fails host validation and crashloops the gateway, so it must be declared in both places.

Build, typecheck and the full test suite pass; adds three `matchesSkipPattern` unit tests (mid-content case-insensitive match, empty/undefined no-op, and that a loose topical mention without the exact marker is not skipped).

### DIFF
--- a/integrations/openclaw/README.md
+++ b/integrations/openclaw/README.md
@@ -167,6 +167,13 @@ As an extra backstop, `min_capture_chars` (env `MEMINI_MIN_CAPTURE_CHARS`, defau
 **0** = off) drops a capture whose stripped user turn is shorter than N characters —
 set it (e.g. `30`) if a gateway still emits short residual-noise turns.
 
+`capture_skip_patterns` (a list of substrings, default empty) drops any turn that
+contains one of them, matched case-insensitively across both the user and assistant
+text. Unlike the `[cron:` / `[Subagent Context]` markers above — which must prefix the
+user turn — these match anywhere, so they catch structured markers an integration emits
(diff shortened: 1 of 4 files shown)


eleboucher:
> _Originally by **eleboucher** on 2026-09-10 07:17 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/266#issuecomment-17513)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #269), which is the source of truth. Further review and merging happen there.

PR #100, opened 2026-09-16
feat: allow fast automatic prompt recall without reranking

phorrr:
On my memini, I only want to use the reranker on explicit memory searches because the reranker introduces a couple of seconds of extra lag. So on every prompt this makes my agent too slow.
I kept the default behavior as-is, just introduced an extra config flag.


eleboucher:
> _Originally by **eleboucher** on 2026-09-19 05:46 UTC — [view on source](https://git.erwanleboucher.dev/eleboucher/memini/issues/287#issuecomment-18365)_

/sync

eleboucher:
Promoted to the canonical Forgejo repository (PR #288), which is the source of truth. Further review and merging happen there.

